PR #12, opened 2020-11-11
Allow talking to UPower on system, not session, bus

### LINKED ISSUE #10: Chromium can't talk to UPower
That's because UPower resides on _system_ bus, while `--talk-name=org.freedesktop.UPower` allows name on _session_ bus.
`--system-talk-name=org.freedesktop.UPower` should be used instead.

wjt:
Fixes #10


andrunko (commented):
Fwiw this LGTM.

refi64:
Merged by hand to avoid extra rebuilds.

Issue #11, opened 2020-11-11
Segfault when attempting to use U2F/WebAuthn devices


wjt:
## Steps to reproduce

1. Browse to https://demo.yubico.com/webauthn-technical/registration
2. Hit `Next`

## Expected results

- If a U2F or other WebAuthn device is attached, browser prompts me to use it
- If not, an error is shown that there are no such devices

## Actual results

```
Thread 1 "chrome" received signal SIGSEGV, Segmentation fault.
0x000055555813624c in ?? ()
(gdb) bt
#0  0x000055555813624c in  ()
#1  0x00007fffffffbff0 in  ()
#2  0x000055555a2a9c06 in operator new(unsigned long) ()
#3  0x00005555589f2073 in  ()
#4  0x0000000000000000 in  ()
```

"Happily" (in terms of being easy to reproduce) this happens regardless of whether I have my U2F dongle plugged in or not.

Unfortunately this means that I cannot sign into my Google account: it has 2-factor authentication configured, with two U2F devices, and so the browser crashes if I try to do so. As a result I can't do my daily work – all of which either uses Google apps, or uses Google authentication – and so can't meaningfully test this Flatpak.

## Chromium version

```
          ID: org.chromium.Chromium
         Ref: app/org.chromium.Chromium/x86_64/beta
        Arch: x86_64
      Branch: beta
     Version: 85.0.4183.121
     License: BSD-3-Clause and LGPL-2.1+ and Apache-2.0 and IJG and MIT and GPL-2.0+ and ISC and OpenSSL and (MPL-1.1 or GPL-2.0 or LGPL-2.0)
      Origin: flathub-beta
  Collection: org.flathub.Beta
Installation: system
   Installed: 259,8 MB
     Runtime: org.freedesktop.Platform/x86_64/20.08
         Sdk: org.freedesktop.Sdk/x86_64/20.08

      Commit: 47645bc1ac4efff9e5618e3baa74faf5dd1d447d554d9e624e0c6eeca60c5389
      Parent: 27e3e36a763329a029aaf9bd6406cf54d6b0c3ef3c36166363b789924dd6b55d
     Subject: Update to Update to 86.0.4240.183 (c88e868b)
        Date: 2020-11-03 03:42:22 +0000
```

Issue #1, opened 2020-10-16
Processes remaining after browser close


refi64:
Likely an issue with the Flatpak sandbox implementation.

x80486:
This is fixed for me...probably a candidate to be closed :wink:

refi64:
Whoops, forgot to close after merge

PR #13, opened 2020-11-12
Update to Chromium 86.0.4240.198 & more patches

refi64:
This adds:
- Experimental desktop shortcut support with flextop
- Widevine extension point tweaks in line with the official ABI
- Tries to fix the hanging process issue
- Fixes chrome://restart


refi64:
Should fix #11, #1.

PR #19, opened 2020-11-18
Update to Chromium 87.0.4280.66 w/ latest patches

### LINKED ISSUE #14: WebAuthn navigator.credentials methods never resolve
When attempting to log into my Google account, which has two-factor authentication enabled and two U2F devices enrolled, I get stuck at the following screen:

Capture d’écran de 2020-11-17 09-43-42

It may not be completely obvious from the screenshot but all elements (except the ones at the bottom, to e.g. switch language and restart the authentication flow) are insensitive. This occurs regardless of whether I have a U2F token connected to my computer. There is no option to use another two-factor authentication mechanism, such as a code from the TOTP generator on my phone. As a result I cannot sign into my Google account (whether for browser synch or for email/calendar/docs/Meet/etc).

On GitHub I instead get an error immediately when I try to use my security key, at which point it offered me the option to use an alternative mechanism, which is how I am writing this comment.

I tried the demo at https://webauthn.io/, which hung in the same way as Google authentication. Stepping through in the debugger (comments inline):

```js
            console.log("Credential Creation Options");
            console.log(makeCredentialOptions);
            navigator.credentials.create({ // <-- this line is executed
                publicKey: makeCredentialOptions.publicKey
            }).then(function (newCredential) {
                console.log("PublicKeyCredential Created"); // <-- but this line is never reached: the promise never resolves
                console.log(newCredential);
                state.createResponse = newCredential;
                registerNewCredential(newCredential);
            }).catch(function (err) {
                console.info(err); // <-- if I try to register a second time while the first operation is pending, the promise errors and this line is reached
            });
```

### LINKED ISSUE #15: Some H.264 videos stutter, even with .Codecs

refi64:
**This has not been fully built and tested locally.** A full build would likely take around an entire day, so I'm pushing this up here a bit early in hopes that any issues can be worked out more quickly.

The following additional changes have been made:

- Widevine is now exposed into the sandbox, so loading the updated
  component variant should now be working. As a result, the global
  widevine extension point has been removed.
- flextop was updated to ff7896f119dce4038044150225fa8404c5607127,
  with the following changes:
    - Only report host access issues on install
  This *should* fix part of #15 (the alert dialog being shown when just
  uninstalling extensions).
- Access is granted by default to ~/.local/share/applications and
  ~/.local/share/icons. The user can still opt to remove access to these
  directories without particularly breaking things. Fixes #15.
- A more proper fix for #11 was implemented, which also fixes #14.
  The stub HidService was failing to resolve several callbacks.
- While solving the above, a more proper fix for #1 was implemented. The
  device manager was never connecting to the Mojo pipe if the USB
  service could not be created, leading to an infinite loop of trying to
  re-create the service.
- An extension point for Chromium extensions has been added.
- The codec loading code was cleaned up and also better reports errors.
  This should help with #15.
- Support for the OpenDirectory method of the OpenURI portal has been
  added. This is *not* thoroughly tested at the moment, due to
  flatpak/xdg-desktop#545.
- The XCURSOR_PATH changes were removed, as upstream now consults
  libXcursor for the cursor path, which will take the Flatpak-specific
  cursor loading code into effect.
- Several backported patches are no longer needed and removed.
- Some new build fixes to remove arguments that only work on LLVM SVN.
- The new PGO builds are temporarily disabled, as the LLVM toolchain
  version in the Flatpak environment cannot read the new PGO data. We
  will likely have to switch to using the Chromium-provided Clang
  toolchain in order to solve this.


Erick555:
> Access is granted by default to ~/.local/share/applications and
> ~/.local/share/icons. The user can still opt to remove access to these
> directories without particularly breaking things. Fixes #15 

That's surprising connection between those paths and stuttering videos 😕 .

> The new PGO builds are temporarily disabled, as the LLVM toolchain
> version in the Flatpak environment cannot read the new PGO data. We
> will likely have to switch to using the Chromium-provided Clang
> toolchain in order to solve this.

Using chromium toolchain sounds a right thing to do even without PGO.

patches/0008-flatpak-Adjust-paths-for-the-sandbox.patch
```diff
@@ -80,7 +80,21 @@ index 26dc3fc1cafa8..8f26e588b4e63 100644
...
  #else
-@@ -521,8 +533,8 @@ bool PathProvider(int key, base::FilePath* result) {
+@@ -511,7 +523,12 @@ bool PathProvider(int key, base::FilePath* result) {
+ #endif
+ #if defined(OS_LINUX) || defined(OS_CHROMEOS)
+     case chrome::DIR_STANDALONE_EXTERNAL_EXTENSIONS: {
+-      cur = base::FilePath(kFilepathSinglePrefExtensions);
++      if (sandbox::FlatpakSandbox::GetInstance()->GetSandboxLevel() >
++          sandbox::FlatpakSandbox::SandboxLevel::kNone) {
++        cur = base::FilePath(FILE_PATH_LITERAL("/app/extensions"));
```
andrunko, line 90: Looks like we also need to adjust the other paths here (e.g. `/app/policies`) with the changes to the extension points dirs.

refi64:
New changes:

- Had to drop using the XDG file chooser portal by default due to flatpak/xdg-desktop-portal#545.
- Policies extension point was changed.
- Flextop was updated to remove the icon prefixes & app subdirectory usage.

Erick555:
> Had to drop using the XDG file chooser portal by default due to flatpak/xdg-desktop-portal#545.

Does this affect chromium specifically or should other flatpaks like ff avoid using portals now?

refi64:
@Erick555 it mostly affects browsers I think because trying to click a downloaded file to open it or using "open in directory" doesn't work.

GeorgesStavracas:
Chromium 87 comes with Ozone/Wayland, so it might be interesting to expose the Wayland socket too.

refi64:
Oh whoops I had that enabled and removed it by accident, will merge by hand to avoid spurious rebuilds

PR #24, opened 2020-11-19
Try enabling new-dtags to fix FFmpeg

### LINKED ISSUE #15: Some H.264 videos stutter, even with .Codecs

refi64:
new-dtags will use DT_RUNPATH instead of DT_RPATH, which means that
LD_LIBRARY_PATH will be given higher priority. This breaks standard
component builds, since DT_RUNPATH isn't used to resolve transitive
dependencies, but that doesn't matter for our static release builds.

Closes #15.

### DIFF
--- a/org.chromium.Chromium.yaml
+++ b/org.chromium.Chromium.yaml
@@ -254,6 +254,7 @@ modules:
           - patches/0019-Add-support-for-the-XDG-file-chooser-portal.patch
           - patches/0020-Add-OpenURI-portal-support-for-opening-directories.patch
           - patches/0021-Remove-references-to-enable-dse-memoryssa.patch
+          - patches/0022-Enable-new-dtags-on-non-component-builds.patch
       - type: file
         path: org.chromium.Chromium.desktop
       - type: file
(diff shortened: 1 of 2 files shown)


refi64:
I just realized, this didn't work because I forgot to apply the patch 😅 will fix tomorrow

andrunko:
Build 31685 still fails for me, here are my findings:
- `RUNPATH` is correctly set in the `chrome` binary while `RPATH` is not, as expected
- `add-ld-path` seems to only affect `ld.so.cache` (not `LD_LIBRARY_PATH` env variable) which is searched after `RUNPATH` - see https://man7.org/linux/man-pages/man8/ld.so.8.html
- Setting `LD_LIBRARY_PATH=/app/chromium/nonfree-codecs/lib:/app/chromium` (or only `LD_LIBRARY_PATH=/app/chromium/nonfree-codecs/lib`) in the `chromium` startup script makes it load both the nonfree and free versions :thinking: (which makes me suspect the sandboxed processes are not inheriting `LD_LIBRARY_PATH`):
```
[org.chromium.Chromium ~]$ chromium --enable-logging=stderr --v=1 2>&1 | grep 264
[165:165:1120/160049.535898:VERBOSE1:mime_util_internal.cc(311)] Using FFmpeg's H.264 decoder
[268:35:1120/160049.630430:VERBOSE1:mime_util_internal.cc(327)] Using Cisco OpenH264
[226:10:1120/160049.633293:VERBOSE1:mime_util_internal.cc(327)] Using Cisco OpenH264
```

So I believe we still need to set `LD_LIBRARY_PATH` manually and understand/fix the issue with the subprocesses (?) preferring the version on `RUNPATH`.

PR #28, opened 2020-11-22
Fix FFmpeg, and various other fixes

### LINKED ISSUE #15: Some H.264 videos stutter, even with .Codecs

### LINKED ISSUE #25: Metainfo showing outdated version; Chromium having access to $HOME
The metainfo is showing an outdated version of Chromium: https://github.com/flathub/org.chromium.Chromium/blob/beta/org.chromium.Chromium.metainfo.xml#L29.

As seen in GNOME Software:
image

---

Another thing is that the flatpak now has access to $HOME: https://github.com/flathub/org.chromium.Chromium/blob/beta/org.chromium.Chromium.yaml#L8. Not sure if it was intended or not, I just wanted to point that out.

refi64:
This does the following:

- Removes new-dtags from non-component builds:
  new-dtags will use DT_RUNPATH instead of DT_RPATH, which means that
  LD_LIBRARY_PATH will be given higher priority. This breaks standard
  component builds, since DT_RUNPATH isn't used to resolve transitive
  dependencies, but that doesn't matter for our static release builds.
  Closes #15.
- Updates the Chromium version history in the metainfo file:
  This had grown pretty grossly out of date. Closes #25.
- Tweaks the extension points (again).

### DIFF
--- a/org.chromium.Chromium.metainfo.xml
+++ b/org.chromium.Chromium.metainfo.xml
@@ -26,6 +26,11 @@
     </screenshot>
   </screenshots>
   <releases>
+    <release version="87.0.4280.66" date="2020-11-18" />
+    <release version="86.0.4240.198" date="2020-11-12" />
+    <release version="86.0.4240.183" date="2020-11-02" />
+    <release version="86.0.4240.111" date="2020-11-01" />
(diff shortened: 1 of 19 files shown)


refi64:
Merged by hand

PR #30, opened 2020-12-04
Update Flextop & Chromium

refi64:


wjt:
I commented on flextop#3 why I am suspicious that the tryexec path will be too specific and so will make the desktop files disappear the moment the app is updated. Not at computer to test right now.

wjt:
Works perfectly.

PR #32, opened 2020-12-05
Update pipewire

julio641742:


refi64:
Some context: it *needs* to be pipewire 0.2, because the ABI changed in 0.3 and Chromium doesn't support the new one yet.

PR #38, opened 2020-12-11
Document extension points

### LINKED ISSUE #37: Add support for policies
Would it be possible to add support for host policies? i.e. by adapting and backporting https://github.com/flathub/com.google.Chrome/commit/7aeccece837f6a838f9eba6001cda8a8ef06db16

andrunko:
Closes #37.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,56 @@
 ## Chromium Flatpak
 
+### Extension points
+
+To avoid having to expose more of the host filesystem in the sandbox but still
+allowing extending Chromium, the following extension points are defined:
+- org.chromium.Chromium.Policy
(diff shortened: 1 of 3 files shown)


refi64:
Note to self: stuff like should probably be in a separate repo / wiki in the future to avoid all the rebuilds (unless we can skip flathub rebuilds via commit message?)

andrunko:
> Note to self: stuff like should probably be in a separate repo / wiki in the future to avoid all the rebuilds (unless we can skip flathub rebuilds via commit message?)

Good point, I was actually surprised when I saw the builds, initially thought they were triggered only for changes to the manifest.

Issue #36, opened 2020-12-07
Unable to use system proxy settings


laolux:
I just installed the chromium flatpak and cannot change the proxy settings.
Opening "settings -> advanced -> Open your computer's proxy settings" does open a new tab which reads

    When running Chromium under a supported desktop environment, the system proxy settings will be used. However, either your system is not supported or there was a problem launching your system configuration.

    But you can still configure via the command line. Please see man chrome for more information on flags and environment variables.

I suppose some permission is missing.

Info about my system: Fedora 33, kernel 5.9.12, flatpak 1.8.3

EDIT: specifying the correct proxy on the command line when starting chromium works:

    flatpak run org.chromium.Chromium --proxy-server=<address:port>

andrunko:
Just to clarify, does chrome respects the settings if you configure via GNOME control center?

As for opening the settings UI from chrome settings, I did a quick check and this is done on `chrome/browser/ui/webui/settings_utils_linux.cc` which launches a system command depending on the environment it is running (GNOME, KDE, etc), so we probably need to special case here when running under flatpak sandbox to run the commands to open the settings via `flatpak-spawn --host` or something similar.

wjt:
https://flatpak.github.io/xdg-desktop-portal/portal-docs.html#gdbus-org.freedesktop.portal.ProxyResolver there is a portal and glib knows how to use it

andrunko:
Yes, but afaics the portal doesn't handle opening the system settings (maybe it should?), and chromium launches the configuration by calling the executables directly. I would be interested to know if chromium actually respects the settings configured though.

laolux:
> Just to clarify, does chromium respects the settings if you configure via GNOME control center?

No, it completely ignores the proxy settings. Using a locally installed chromium works out of the box, so it must be the flatpak that is ignoring the settings from the GNOME control center.

> As for opening the settings UI from chromium settings, I did a quick check and this is done on `chrome/browser/ui/webui/settings_utils_linux.cc` which launches a system command depending on the environment it is running (GNOME, KDE, etc), so we probably need to special case here when running under flatpak sandbox to run the commands to open the settings via `flatpak-spawn --host` or something similar.

yes, when I try to open the proxy settings I get a warning message in my terminal window:

    [2:2:1208/123458.525340:ERROR:settings_utils_linux.cc(54)] Could not find GNOME network settings in $PATH

andrunko:
[...]

PR #41, opened 2020-12-17
Add support for respecting system proxy settings

andrunko:
Ok, this is an attempt to implement support for respecting system proxy settings on the flatpak version.

Some general notes on implementation:
- The implementation uses the same codepath used when `--winhttp-proxy-resolver` is passed as param, which uses the `net::ProxyResolver` interface to determine the proxy given an url:
  - This is due to limitations in the portal APIs which don't allow access to the system proxy settings, and with that no reliable way to implement a proper ProxyConfigService when on flatpak
  - Despite the name of the param, this flag is not Windows(tm) specific and also used on Mac builds
  - This param may be removed at some point in the future, see https://bugs.chromium.org/p/chromium/issues/detail?id=644030 but it should be replaced with https://bugs.chromium.org/p/chromium/issues/detail?id=1032820, which from the looks of it would not require too many changes to the impl here
  - To avoid having to pass this param I've enabled this codepath automatically when detecting we are running on flatpak
- This means that proxy resolution is done by invoking `net::ProxyResolver::GetProxyForURL()` (instead of relying on the internal chromium mechanism based on the proxy settings, set to "mode=auto" in this case), which in turn invokes `g_proxy_resolver_lookup` which uses the portal API via D-Bus
  - To avoid having one D-Bus call per `net::ProxyResolver::GetProxyForURL()`, I've implemented a (very) simple cache that gets invalidated every 1min (same default timeout used on chromium dns refresh)
- Another issue is that the impl will invoke the portal even if proxy is disabled, also because of a limitation in the portal API where there is no way to know whether the proxy is disabled
- We could skip using the portal on KDE, given we have access to $HOME today (and the impl just reads .ini files), but I decided to go with the same impl when running on flatpak, but have no hard-preference here

On the bright side, it seems to work fine from my tests (could use some more tests though) and changes to the system settings (tested on GNOME/Endless) are reflected correctly (considering the cache invalidation timeout of 1min).

I am not really happy with the amount of D-Bus calls trigerred here tbh (esp. when proxy is disabled), but I can't think of another way to implement this with the current portal APIs - we could tweak the cache invalidation timeout but apart from that I don't think we could do much to improve.

See https://github.com/flatpak/xdg-desktop-portal/issues/554 for the portal API issues/limitations mentioned above.

Thoughts?

Closes https://github.com/flathub/org.chromium.Chromium/issues/36

### DIFF
--- a/patches/0023-Add-support-for-respecting-system-proxy-settings-whe.patch
+++ b/patches/0023-Add-support-for-respecting-system-proxy-settings-whe.patch
@@ -1,0 +1,388 @@
+From 2fc65c6722cdd50accf044868de3a4614309c172 Mon Sep 17 00:00:00 2001
+From: Andre Moreira Magalhaes <andre.magalhaes@endlessos.org>
+Date: Thu, 17 Dec 2020 19:05:10 -0300
+Subject: [PATCH 23/23] Add support for respecting system proxy settings when
+ running on Linux/Flatpak
+
+Some general notes on implementation:
+- The implementation uses the same codepath used when
+  `--winhttp-proxy-resolver` is passed as param, which uses the
+  `net::ProxyResolver` interface to determine the proxy given an url:
+  - This is due to limitations in the portal APIs which don't allow
+    access to the system proxy settings, and with that no reliable
+    way to implement a proper ProxyConfigService when on flatpak
+  - Despite the name of the param, this flag is not Windows(tm)
+    specific and also used on Mac builds
+  - This param may be removed at some point in the future, see
+    https://bugs.chromium.org/p/chromium/issues/detail?id=644030 but
+    it should be replaced with
+    https://bugs.chromium.org/p/chromium/issues/detail?id=1032820,
+    which from the looks of it would not require too many changes
+    to the impl here
+  - To avoid having to pass this param, this codepath is enabled
+     automatically when detecting we are running on flatpak
+- This means that proxy resolution is done by invoking
+  `net::ProxyResolver::GetProxyForURL()` (instead of relying on the
+  internal chromium mechanism based on the proxy settings, set to
+  "mode=auto" in this case), which in turn invokes
+  `g_proxy_resolver_lookup` which uses the portal API via D-Bus
+  - To avoid having one D-Bus call per
+    `net::ProxyResolver::GetProxyForURL()`, this implements a (very)
+    simple cache that gets invalidated every 1min (same default
+    timeout used on chromium dns refresh)
+- Another issue is that the impl will invoke the portal even if
+  proxy is disabled in the system, also because of a limitation
+  in the portal API where there is no way to know whether the
+  proxy is disabled
+- We could skip using the portal on KDE, given the flatpak has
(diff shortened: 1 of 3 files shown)


patches/0023-Add-support-for-respecting-system-proxy-settings-whe.patch
```diff
@@ -0,0 +1,504 @@
...
++#include <gio/gio.h>
++#endif  // defined(USE_GIO)
++
++namespace net {
++
++class NetworkIsolationKey;
++
++namespace {
++
++class GLibThread {
```
refi64, line 260: Question: do we actually need a separate thread here? GLib declares global structures to be thread-safe, so I believe `g_proxy_resolver_lookup_async` could be called from the same thread, locking the cache if necessary.

andrunko: Tbh I wasn't sure either and on a **quick** look at the code (portal impls) they didn't look thread safe to me so I went with the safest option. Maybe @pwithnall would be able to confirm/clarify here.

pwithnall: @refi64 is right: you should be able to expect that the proxy resolver returned by `g_proxy_resolver_get_default()` is thread-safe.

The `gsimpleproxyresolver.c` implementation is thread-safe as long as there are no calls to `g_simple_proxy_resolver_set_*()` after any threads have been spawned.

The `gproxyresolverportal.c` implementation is entirely thread-safe, because it just passes calls through to a `GDBusProxy` around the portal, which is thread-safe.

The `gproxyresolvergnome.c` implementation (in glib-networking) looks like it’s thread-safe (it has a mutex), but I haven’t checked it thoroughly.

The `glibproxyresolver.c` (libproxy) implementation (in glib-networking) also looks like it has thought about threading, because it already does its libproxy calls in a worker thread.

tl;dr: You should be able to drop your worker thread here. If any threading bugs do appear, they are likely bugs in GLib; but I wouldn’t expect any to appear if you’re using any implementation except `gsimpleproxyresolver.c`. It would be unlikely you’d end up using that implementation, as it’s a fallback.

andrunko: Thanks @pwithnall, @refi64. I've updated the impl to drop the glib worker thread. Some possible improvements using current GProxyResolver/Portal interface APIs (for the future - I don't plan to work on them for the time being):
- avoid calls to `g_proxy_resolver_lookup()` if there is an existing inflight request for the same url
  - although `net::ProxyResolver::GetProxyForURL()` is always invoked from the same thread for the same `net::ProxyResolver` instance, different instances may be invoked from different threads but still use the same `GProxyResolver` instance...
- using a single/global cache for all proxy resolver instances
  - as above, given we use the same `GProxyResolver` instance behind the sceces, we could consider a global cache
- make cache smarter
- allow async `net::ProxyResolver::GetProxyForURL()` calls (although currently the only consumer is `net::MultiThreadProxyResolver` which always expects a sync call)

Issue #29, opened 2020-11-26
Wire up flatpak-external-data-checker


wjt:
Chromium releases very frequently: new major releases every 6 weeks, and there are often 1-3 patches in between major releases which fix major security issues.

https://github.com/flathub/flatpak-external-data-checker exists to automatically open PRs for new releases.

Whether by adding a Chromium-specific checker or by wiring up one of the existing scrapers, we should arrange for flatpak-external-data-checker to open PRs for this repo. (@andrunko please correct me if I'm wrong, but) the patches Endless OS's built-in chromium package carries would typically apply cleanly to point releases, but often needed some minor refreshments to apply to new major versions. I still think the failed PR build for a new major version is useful as a reminder to do that refreshment.

andrunko:
> @andrunko please correct me if I'm wrong, but) the patches Endless OS's built-in chromium package carries would typically apply cleanly to point releases, but often needed some minor refreshments to apply to new major versions. I still think the failed PR build for a new major version is useful as a reminder to do that refreshment.

Yes, that is correct, I also agree having the PR would help even if it fails to build (if only to notify when there is a new upstream release available).

refi64:
I'm not entirely sure how this would be wired up without adding some extra
stuff to the external data checker first. It often take a few hours for the
source tarballs to go up after pretty much all release announcements to go
up, so the main other way would be go query the Cloud Storage APIs for new
tarballs, *but* that doesn't seem to provide any way to distinguish stable
releases so that would have to be done some other way...

gasinvein:
Can we use https://omahaproxy.appspot.com? I'm not sure if it's the official way to check for Chromium releases, but it seems to be what people are using.

andrunko:
Yes, I think so, it also supports filtering the json: https://omahaproxy.appspot.com/all.json?os=linux;channel=stable.

PR #47, opened 2021-01-06
Plug external data checker

andrunko:
This uses the anitya checker to monitor updates:
- https://release-monitoring.org/project/13344/

Fixes: https://github.com/flathub/org.chromium.Chromium/issues/29

### DIFF
--- a/org.chromium.Chromium.metainfo.xml
+++ b/org.chromium.Chromium.metainfo.xml
@@ -1,5 +1,5 @@
-<!-- Copyright 2017 The Chromium Authors -->
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2017 The Chromium Authors -->
 <component type="desktop">
   <id>org.chromium.Chromium</id>
   <launchable type="desktop-id">org.chromium.Chromium.desktop</launchable>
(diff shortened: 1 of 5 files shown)


andrunko:
Note that this will very likely produce PRs with conflicts when updating major versions, but should work fine in most cases at least for minor updates.

Some of the changes could arguably be done in flatpak-external-data-checker itself (e.g. to avoid rewriting coding style), but they are simple enough that I don't think it is worth it.

org.chromium.Chromium.metainfo.xml
```diff
@@ -1,4 +1,3 @@
-<!-- Copyright 2017 The Chromium Authors -->
```
wjt, line 1: You can put it after the `<?xml ...?>` tag. I went to [great lengths](https://github.com/flathub/flatpak-external-data-checker/blob/8b5dfa2891b10502cc63804248634e1213a41396/src/lib/appdata.py#L92-L125) to make this work. [Here's the test](https://github.com/flathub/flatpak-external-data-checker/blob/8b5dfa2891b10502cc63804248634e1213a41396/tests/test_appdata.py#L192-L223).

wjt (changes requested):
Apart from the build failure which looks to be related to moving a script out to an actual file, and one nitpick, this LGTM

wjt:
> Some of the changes could arguably be done in flatpak-external-data-checker itself (e.g. to avoid rewriting coding style), but they are simple enough that I don't think it is worth it.

I went to some lengths to preserve YAML formatting, [e.g.](https://github.com/flathub/flatpak-external-data-checker/blob/8b5dfa2891b10502cc63804248634e1213a41396/src/lib/utils.py#L280-L284), but reached a point where I couldn't go much further without disproportionate effort :-)

org.chromium.Chromium.metainfo.xml
andrunko: Thanks, updated - although not sure how accurate this copyright notice is.

andrunko:
Pushed some changes to try fixing the build, let's see (I was testing the scripts manually but forgot to test integration with the manifest...).

org.chromium.Chromium.metainfo.xml
wjt: Agreed. But at least it's still here so it could be updated :-)

wjt:
I tested this locally, and I got:

```
+ 2021-01-06 22:31:30,090 WARNING src.checkers.htmlchecker: https://commondatastorage.googleapis.com/chromium-browser-official/chromium-87.0.4280.141.tar.xz returned HTTP Error 404: Not Found
BROKEN: chromium-87.0.4280.88.tar.xz
 Unreachable URL: https://commondatastorage.googleapis.com/chromium-browser-official/chromium-87.0.4280.88.tar.xz
```

It seems the new tarball hasn't appeared yet, but also that something weird is going on with the old URL. Maybe it's just a bug in the checker. Presumably once the new tarball appears the checker would succeed.

andrunko:
> I tested this locally, and I got:
> 
> ```
> + 2021-01-06 22:31:30,090 WARNING src.checkers.htmlchecker: https://commondatastorage.googleapis.com/chromium-browser-official/chromium-87.0.4280.141.tar.xz returned HTTP Error 404: Not Found
> BROKEN: chromium-87.0.4280.88.tar.xz
>  Unreachable URL: https://commondatastorage.googleapis.com/chromium-browser-official/chromium-87.0.4280.88.tar.xz
> ```
> 
> It seems the new tarball hasn't appeared yet, but also that something weird is going on with the old URL. Maybe it's just a bug in the checker. Presumably once the new tarball appears the checker would succeed.

Yes, the new tarball hasn't appeared yet indeed (funny that I tested earlier and there was no `.141` announced yet). 
The error seems to be an issue with the `Alitya` checker that inherits from `HTMLChecker` and uses `HTMLChecker._update_version()` to validate the new URL which then sets it to `BROKEN` as it can't fetch it (404). The debug output is wrong because the summary uses the URL for the current version for `BROKEN` data, but it should be a non-issue.

I guess we have to live with the fact that release-monitoring may get updated before the new release tarball is uploaded.

refi64:
New version is up so I guess it's the perfect time to take this for a spin, any issues in the actual checker of course aren't critical to this itself

PR #55, opened 2021-01-20
Update Chromium to 88.0.4324.96

### LINKED ISSUE #35: Refresh VAAPI patches
Upstream Chromium:
* added "chrome flag" for VAAPI decoding in version 86 (#enable-accelerated-video-decode) . Commit ['Allow the flag disable-accelerated-video-decode to be disabled on linux'](https://github.com/chromium/chromium/commit/a59445228676c7f517905f896ee71eaf3741d4bc#diff-b98ee78788d24ecc151c3f8e74fef65be502cb8f1e28fe591d5cfb54010c8a96)
* Set "use_vaapi" flag by default in version 88 (required for "chrome flag" to work). Commit ['Reland "Enable vaapi decoding on linux x86/64 by default." '](https://github.com/chromium/chromium/commit/d658439002fc9bc6b4deeec71738600402ea9742#diff-fa163b664861acb517b8a57ea9fda19f2763df562e38d4d82de346b0ddbea810)
That means that with version 88 we'll probably have VAAPI support out of box for most Chromium builds.

So my proposes for current version 87:
* Drop patch "0013-Enable-VAVDA-VAVEA-and-VAJDA-on-linux-with-VAAPI-onl.patch". It's large and obsolete now.
* Integrate [Fedora patches](https://src.fedoraproject.org/rpms/chromium/tree/master) to make VAAPI experience even better ( specifically [chromium-88-vaapi-attribute.patch](https://src.fedoraproject.org/rpms/chromium/raw/master/f/chromium-88-vaapi-attribute.patch) , [chromium-86.0.4240.75-fix-vaapi-on-intel.patch](https://src.fedoraproject.org/rpms/chromium/raw/master/f/chromium-86.0.4240.75-fix-vaapi-on-intel.patch) , [chromium-86.0.4240.75-enable-hardware-accelerated-mjpeg.patch](https://src.fedoraproject.org/rpms/chromium/raw/master/f/chromium-86.0.4240.75-enable-hardware-accelerated-mjpeg.patch) )

refi64:
**A full build has not yet been tested locally!**

This also contains the following changes:

- The FFmpeg build configuration is now re-generated from scratch,
  rather than manually merging.
  - Note that the OpenH264 configuration was accidentally included with
    the fdk-aac patch. This will be corrected in a future patch set.
  - Only the x64 and arm64 configuration was changed, since those are
    the only ones we build.
  - Several build scripts to make this easier are in the utils/ffmpeg
    directory.
- The VA-API patches were removed in favor of a heavily slimmed down
  patch that enables hardware-accelerated mjpeg. From what I can tell,
  the NVIDIA/VDPAU support the original patches may have had was lost
  long ago, likely amidst a merge. Closes #35.
- Another backported patch "Remove references to enable-dse-memoryssa"
  was removed, as its contents are now present in Chromium 88.
- The error message for when unprivileged user namespaces are not
  available has been rewritten for simplicity.


refi64:
So we *did* still need a modified version of the enable-dse-memoryssa patch, it's been re-added back now.

refi64:
Welp, this seems to have AAC embedded in the FFmpeg binaries somehow, I'll have to take a look at those patches again...

patches/0001-ffmpeg-Use-royalty-free-libfdk-aac-for-AAC-playback.patch
```diff
@@ -587,23 +1700,43 @@ index 1908ba19e..920b22bfa 100644
...
-+        # Build against libfdk-aac and enable it always.
++    # Depend on openh264 for H.264 support.
++    configure_flags['Chromium'].extend([
++        '--enable-libopenh264',
++        '--enable-decoder=libopenh264',
++        '--disable-decoder=h264',
++    ])
++
++    # Depend on royalty-free libfdk-aac for AAC support.
++    configure_flags['Chrome'].extend([
```
andrunko, line 1731: I don't think we want that, we want fdk-aac enabled in the free (`Chromium`) configuration at least - actually thinking about it I think we want fdk-aac in the Chromium config and use the built-in AAC decoder (ff_aac_decoder) in the Chrome (non-free) config? As is now (current master) both configurations are using fdk-aac.

andrunko: fdk-aac only supports a subset of AAC profiles so I think we should do the same as we do for openh264 and only enable it in the free configuration.

andrunko: (fwiw, the behaviour suggested above is actually what our current tests expect on Endless OS).

refi64: @andrunko I don't think that's correct, the fdk-aac license says

> NO EXPRESS OR IMPLIED LICENSES TO ANY PATENT CLAIMS, including without limitation the patents of Fraunhofer, ARE GRANTED BY THIS SOFTWARE LICENSE. Fraunhofer provides no warranty of patent non-infringement with respect to this software.
>
> You may use this FDK AAC Codec software or modifications thereto only for purposes that are authorized by appropriate patent licenses.

which probably shouldn't be in the free builds?

refi64: It's also mentioned [on the FFmpeg page](https://trac.ffmpeg.org/wiki/Encode/AAC):

> The license of libfdk_aac is not compatible with GPL, so the GPL does not permit distribution of binaries containing incompatible code when GPL-licensed code is also included. Therefore this encoder have been designated as "non-free", and you cannot download a pre-built ffmpeg that supports it.

refi64: and in their news page

> Licensing has always been an issue with encoding AAC audio as most of the encoders have had a license making FFmpeg unredistributable if compiled with support for them. The fact that there now exists a fully open and truly free AAC encoder integrated directly within the project means a lot to those who wish to use accepted and widespread standards.

Maybe that was just backwards?

andrunko: Aiui we want to enable fdk-aac in the free configuration but I don't see a point in enabling it in the non-free configuration. Fwiw, libfdk-aac is shipped in the `org.freedesktop.Platform` runtime. 

In any case we can do this change (to only enable fdk-aac in the free config) in a separate PR if you prefer, but we should at least enable it in the free builds (current status quo) for now to avoid regressions.

If we make this change to only enable fdk-aac in the free config we will also need to adjust `CheckNonFreeMimeTypesOnFlatpak()` on `media/base/mime_util_internal.cc` to check for `ff_aac_decoder` when deciding if AAC is supported (+ adjust the commit msg for that patch).

refi64:
FWIW regardless of the result of the previous discussion, this current configuration should mimic the previous one as closely as possible other than the fdk-aac misstep which is now fixed, since we really need M88 in the repos.

andrunko:
> FWIW regardless of the result of the previous discussion, this current configuration should mimic the previous one as closely as possible other than the fdk-aac misstep which is now fixed, since we really need M88 in the repos.

The "only" diff now (as mentioned in a prev comment) is that we are enabling the AAC parser on free builds as well, but I think that is fine, afaiu the parser doesn't have any licensing issues and tbh I think not enabling it before was an oversight.

refi64:
That should be perfect then.

patches/0001-ffmpeg-Use-royalty-free-libfdk-aac-for-AAC-playback.patch
andrunko: @refi64 Just to follow-up, looks like we still want to keep fdk-aac for non-free builds (as well as free builds) as we still plan to ship the `.Codecs` extension on some EOS images without incurring AAC royalties. @ramcq @wjt is that correct?

wjt: Yes, please can we keep only patent-unencumbered fdk-aac in `.Codecs` too, and not a patent-encumbered AAC implementation?

PR #68, opened 2021-02-20
desktop: Add 'chrome' and 'internet' keywords

wjt:
@barthalion said:

> looking at some bare search stats, you may want to add "chrome" to the
> keywords in chromium's appdata file

I tried a few other obvious keywords and "internet" is the other one
which doesn't match.

(When the AppStream Collection Metadata is created, [keywords from `.desktop` files are extracted](https://www.freedesktop.org/software/appstream/docs/chap-CollectionData.html#tag-ct-keywords).)

### DIFF
--- a/org.chromium.Chromium.desktop
+++ b/org.chromium.Chromium.desktop
@@ -112,6 +112,7 @@ Terminal=false
 Icon=org.chromium.Chromium
 Type=Application
 Categories=Network;WebBrowser;
+Keywords=chrome;internet;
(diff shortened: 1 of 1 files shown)


wjt:
I killed that build because it is a waste of electrons to compile a whole browser for this.

```
wjt@camille:~$ desktop-file-validate src/flathub/org.chromium.Chromium/org.chromium.Chromium.desktop; echo $?
0
```

refi64:
Merged by hand.

PR #73, opened 2021-03-04
Configure flatpak-external-data-checker to monitor stable releases only

andrunko:
This is needed due to a recent change to the anitya checker:
  https://github.com/flathub/flatpak-external-data-checker/commit/12825bfd1de6669d32006d201ef1f827e43575ff

### DIFF
--- a/org.chromium.Chromium.yaml
+++ b/org.chromium.Chromium.yaml
@@ -189,6 +189,7 @@ modules:
         x-checker-data:
           type: anitya
           project-id: 13344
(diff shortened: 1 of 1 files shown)


andrunko:
> Build [40879 was cancelled](https://flathub.org/builds/#/builders/32/builds/40879)

Cancelled as there is no point spending CPU cycles for this change.

wjt:
Thanks, I also cancelled the real build.

PR #79, opened 2021-03-09
Add pcsc socket permission

### LINKED ISSUE #189: YubiKey FIDO2 WebAuthn requires pcsc socket
WebAuthn was failing for me, falling back to OTP.

As soon as I added the pcsc socket permission, voila.

iavael:
Fixes #189

### DIFF
--- a/org.chromium.Chromium.yaml
+++ b/org.chromium.Chromium.yaml
@@ -14,6 +14,7 @@ finish-args:
   - --share=ipc
   - --share=network
   - --socket=cups
+  - --socket=pcsc
   - --socket=pulseaudio
   - --socket=x11
   - --socket=wayland


refi64:
Do you have an example of what exactly this fixes? I'm not aware of any code in Chromium that tries to directly talk to these sockets.

iavael:
That's needed for communication with pkcs11-compatible smartcards. It's also needed to be able to load pkcs11 modules for specific cards, but it seems to be handlable with flatpak extensions.

refi64:
pcsc does not have a stable protocol, so I don't think this is safe to be merged as-is, see [here](https://github.com/flatpak/flatpak/issues/4723) for more discussion on the issue.

PR #84, opened 2021-03-19
Use llvm11 extension for build

Erick555:
This takes [latest llvm/clang toolchain](https://github.com/flathub/org.freedesktop.Sdk.Extension.llvm11) into use which also allows doing tests like [ASAN](https://github.com/flathub/org.chromium.Chromium/pull/6) without rebuilding llvm.

### DIFF
--- a/build-aux/build.sh
+++ b/build-aux/build.sh
@@ -4,4 +4,5 @@ ninja -C out/ReleaseFree -j$FLATPAK_BUILDER_N_JOBS libffmpeg.so
 
 . /usr/lib/sdk/node12/enable.sh
 . /usr/lib/sdk/openjdk11/enable.sh
+. /usr/lib/sdk/llvm11/enable.sh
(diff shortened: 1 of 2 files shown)


refi64:
Anyone want to test the binary before merge to make sure there are no weird issues?

Also yeah it's gonna be really great to be able to use ASAN here! We might also be able to drop some of the patches to remove newer LLVM flags...

Erick555:
@refi64 the apps started fine for me after those changes but I don't know how to test it more while compiler changes are rather invisible for daily use. The one thing I wish is to have a way to double check that right compiler was used as the build system doesn't provide such info.

refi64:
@Erick555 try a `grep -A4 'rule cxx' out/Default/toolchain.ninja` during build time to check the C++ compiler rule and/or a `which clang++` call

Erick555:
@refi64 I meant a way to print such info in buildlog. Building chromium locally isn't something my system is capable of.

PR #85, opened 2021-03-20
[WIP] Don't replace default CFLAGS during build

Erick555:
Setting C/XXFLAGS env var explicitly replaces default set inherited from runtime which means losing all hardenings done there. Passing additional flags through dedicated f-b option will append them to default set.

### DIFF
--- a/build-aux/build.sh
+++ b/build-aux/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-ninja -C out/ReleaseFree -j$FLATPAK_BUILDER_N_JOBS libffmpeg.so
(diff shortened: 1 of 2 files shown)


refi64:
I believe Chromium should already be passing various hardening flags that are already used by upstream. Since it prefers building with Clang, which has slightly different options, and also has a tendency to find compiler bugs, I'm not super eager to add various other flags that aren't explicitly tested.

That being said, it's not clear to me why this build is failing tbh.

Erick555:
Is there a config option to mae chromium build more verbose? Current output is very cryptic.

PR #86, opened 2021-03-24
Refresh tab-discarding patch description

wjt:
This no longer uses Endless OS's memory pressure implementation.


refi64 (approved):
LGTM, will merge with next batch of updates.

refi64:
Merging in as part of #92.

PR #88, opened 2021-04-06
Update to PipeWire 0.3

grulja:


grulja:
bot, build

tinywrkb:
You have 0.3.21 in the runtime.

```
$ flatpak run --command=pipewire-media-session org.freedesktop.Platform//20.08 --version

pipewire-media-session
Compiled with libpipewire 0.3.21
Linked with libpipewire 0.3.21


$  flatpak run --command=pipewire-media-session org.freedesktop.Sdk//20.08

flatpak run --command=cat org.freedesktop.Sdk//20.08 //usr/lib/x86_64-linux-gnu/pkgconfig/libpipewire-0.3.pc
prefix=/usr
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

moduledir=${libdir}/pipewire-0.3

Name: libpipewire
Description: PipeWire Interface
Version: 0.3.21
Requires: libspa-0.2
Libs: -L${libdir} -lpipewire-0.3
Cflags: -I${includedir}/pipewire-0.3 -D_REENTRANT

```

grulja:
> You have 0.3.21 in the runtime.
> 
> ```
> $ flatpak run --command=pipewire-media-session org.freedesktop.Platform//20.08 --version
> 
> pipewire-media-session
> Compiled with libpipewire 0.3.21
> Linked with libpipewire 0.3.21
> 
> 
> $  flatpak run --command=pipewire-media-session org.freedesktop.Sdk//20.08
> 
> flatpak run --command=cat org.freedesktop.Sdk//20.08 //usr/lib/x86_64-linux-gnu/pkgconfig/libpipewire-0.3.pc
> prefix=/usr
> libdir=${prefix}/lib/x86_64-linux-gnu
> includedir=${prefix}/include
> 
> moduledir=${libdir}/pipewire-0.3
> 
> Name: libpipewire
> Description: PipeWire Interface
> Version: 0.3.21
> Requires: libspa-0.2
> Libs: -L${libdir} -lpipewire-0.3
> Cflags: -I${includedir}/pipewire-0.3 -D_REENTRANT
> ```

Oh, you are right, I didn't realize that. Thank you.

grulja:
bot, build

grulja:
Just for the record, I tested this build and I can correctly share my screen.

Erick555:
What happens if you add `--filesystem=host-etc` permission?

grulja:
It doesn't make any difference.

st 7. 4. 2021 v 15:41 odesílatel Erick555 ***@***.***> napsal:

tinywrkb:
@grulja this should really be fixed in the runtime but you can probably ship the configs and put them in `/app/etc/pipewire` and then set `--env=PIPEWIRE_CONFIG_DIR=/app/etc/pipewire`.  
PipeWire will also pick up configs from `XDG_CONFIG_HOME/pipewire` but by default PipeWire just put the session state there so unless the user copy the configs it won't work.

refi64:
Hmm I'll probably just wait it out until it's fixed in the runtime then to merge, given the importance of a browser & stuff like screen sharing nowadays, vendoring configs feels rather risky in the worst-case scenario.

Erick555:
> PipeWire minor releases might not be so stable with regards to configs so even if you go with the PIPEWIRE_CONFIG_DIR solution, it might break if the runtime was updated with a new PW version but your shipped configs weren't.

Hm, can it work at all when pipewire daemon runs on host and it will almost never match runtime config version?

refi64:
Afaik the configs need to match the library, not necessarily the daemon.

grulja:
bot, build

grulja:
I can confirm that PipeWire 0.3 support works now after the issue has been fixed in Freedesktop runtime.

grulja:
@refi64 can you please re-add the "merge with next release" label since the issue in the runtime has been fixed?

refi64:
Good point, thanks for the reminder.

refi64:
Merging in as part of #92, thanks!

YaLTeR:
This is supposed to be on current Flathub, right? Because I'm getting the crash, which didn't happen when I was testing this particular PR.
```
#0  __memmove_sse2_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:283
#1  0x0000563592da9adb in memcpy (__dest=0x3ee800004000, __src=0x7fd85e5cf000, __len=10240) at /usr/include/x86_64-linux-gnu/bits/string_fortified.h:34
#2  webrtc::BaseCapturerPipeWire::HandleBuffer(pw_buffer*) (this=0x3eec010372c0, buffer=0x3eec0266f310) at ../../third_party/webrtc/modules/desktop_capture/linux/base_capturer_pipewire.cc:858
#3  webrtc::BaseCapturerPipeWire::OnStreamProcess(void*) (data=0x3eec010372c0) at ../../third_party/webrtc/modules/desktop_capture/linux/base_capturer_pipewire.cc:402
#4  0x00007fd879b0f3c0 in do_call_process (loop=<optimized out>, async=<optimized out>, seq=<optimized out>, data=<optimized out>, size=<optimized out>, user_data=<optimized out>) at ../src/pipewire/stream.c:346
#5  0x00007fd863811806 in flush_items (impl=0x3eec01160028) at ../spa/plugins/support/loop.c:141
#6  0x00007fd863811712 in source_event_func (source=0x3eec01c935a0) at ../spa/plugins/support/loop.c:463
#7  0x00007fd863812063 in loop_iterate (object=0x3eec01160028, timeout=<optimized out>) at ../spa/plugins/support/loop.c:308
#8  0x00007fd879b1bb7a in do_loop (user_data=0x3eec01761200) at ../src/pipewire/thread-loop.c:243
#9  0x00007fd87a59d4d2 in start_thread (arg=<optimized out>) at pthread_create.c:477
#10 0x00007fd878adf323 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

grulja:
> This is supposed to be on current Flathub, right? Because I'm getting the crash, which didn't happen when I was testing this particular PR.
> 
> ```
> #0  __memmove_sse2_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:283
> #1  0x0000563592da9adb in memcpy (__dest=0x3ee800004000, __src=0x7fd85e5cf000, __len=10240) at /usr/include/x86_64-linux-gnu/bits/string_fortified.h:34
> #2  webrtc::BaseCapturerPipeWire::HandleBuffer(pw_buffer*) (this=0x3eec010372c0, buffer=0x3eec0266f310) at ../../third_party/webrtc/modules/desktop_capture/linux/base_capturer_pipewire.cc:858
> #3  webrtc::BaseCapturerPipeWire::OnStreamProcess(void*) (data=0x3eec010372c0) at ../../third_party/webrtc/modules/desktop_capture/linux/base_capturer_pipewire.cc:402
> #4  0x00007fd879b0f3c0 in do_call_process (loop=<optimized out>, async=<optimized out>, seq=<optimized out>, data=<optimized out>, size=<optimized out>, user_data=<optimized out>) at ../src/pipewire/stream.c:346
> #5  0x00007fd863811806 in flush_items (impl=0x3eec01160028) at ../spa/plugins/support/loop.c:141
> #6  0x00007fd863811712 in source_event_func (source=0x3eec01c935a0) at ../spa/plugins/support/loop.c:463
> #7  0x00007fd863812063 in loop_iterate (object=0x3eec01160028, timeout=<optimized out>) at ../spa/plugins/support/loop.c:308
> #8  0x00007fd879b1bb7a in do_loop (user_data=0x3eec01761200) at ../src/pipewire/thread-loop.c:243
> #9  0x00007fd87a59d4d2 in start_thread (arg=<optimized out>) at pthread_create.c:477
> #10 0x00007fd878adf323 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
> ```

I think the updated Chromium hasn't been built yet so you still use PipeWire 0.2.

YaLTeR:
Says I'm using 90.0.4430.72, which is what the PR linked above supposed to be if I understand correctly.

grulja:
> Says I'm using 90.0.4430.72, which is what the PR linked above supposed to be if I understand correctly.

You are right.

Can you run these two commands in terminal, each of them separately and let them keep running while trying to reproduce the crash? Then attach please output from both here.

1) `pw-mon`
2) `dbus-monitor --session`

YaLTeR:
[pw-mon.log](https://github.com/flathub/org.chromium.Chromium/files/6333784/pw-mon.log)

[dbus-monitor.log](https://github.com/flathub/org.chromium.Chromium/files/6333785/dbus-monitor.log)

grulja:
> [pw-mon.log](https://github.com/flathub/org.chromium.Chromium/files/6333784/pw-mon.log)
> 
> [dbus-monitor.log](https://github.com/flathub/org.chromium.Chromium/files/6333785/dbus-monitor.log)

This all look correct and I have no idea why this crash happens and it will be hard to debug it since I cannot reproduce it. It would be good to find out when it started crashing and what has changed since then. But I will keep investigating.

refi64:
So it does look like the changeset to use 0.3 by default is in 90, so we can't just revert... I'll see if I can reproduce this, @YaLTeR what pipewire version is on your host, and what distro are you on?

YaLTeR:
F34 Silverblue which currently has 0.3.24.

refi64:
Oh sweet I have the same distro, hopefully I'll be able to reproduce this then.

grulja:
@refi64 if I remember correctly, @YaLTeR had this crash even with PipeWire 0.2 so this is a common issue.

refi64:
So I was able to successfully get a friend to reproduce this on an AMD GPU with experimental DMA-BUF screen sharing enabled in GNOME. @YaLTeR do you share any of that setup?

In our core dump (which looked identical), it turns out this happens on the DMA-BUF path, and somehow the `ScopedBuf` instance ends up with an invalid pointer. Still looking further into it...

YaLTeR:
> So I was able to successfully get a friend to reproduce this on an AMD GPU with experimental DMA-BUF screen sharing enabled in GNOME. @YaLTeR do you share any of that setup?

Indeed, that's exactly my setup. I frequently use OBS for recording and streaming and experimental DMA-BUF helps there.

grulja:
At least we have finally identified the cause. When I enable DMA-BUF it still works for me, but I have an Intel graphics card. If this happens with AMD only, then it might be a driver issue.

YaLTeR:
> At least we have finally identified the cause. When I enable DMA-BUF it still works for me, but I have an Intel graphics card. If this happens with AMD only, then it might be a driver issue.

I believe dma-buf might be unconditionally enabled on Intel because the memory management there, being an integrated GPU, is sightly different.

refi64:
A bit more info I've found: this is not related to the mmap result being invalid, it turns out gdb cannot show mmap'd memory from a core dump.

The part this crashes in glibc's memmove is when copying bytes in 16-byte chunks. In particular, reading the first two chunks is fine, but one of the third or fourth (so either 48 or 64 bytes into the mmap'd memory) results in a SIGBUS. As per the mmap man page, this essentially would happen if the size passed to `mmap` is bigger than the actual size of the fd, and then attempts are made to read into this additional memory.

This is definitely a bit...odd, because from what I can tell, it basically means that the DMA-BUF fd is only a few dozen bytes large, despite having an advertised size of around 8MB? I'm not entirely sure how this would end up happening, but it's definitely at a level I don't think we can really control from Chromium.

refi64:
This may actually be related to: https://gitlab.gnome.org/GNOME/mutter/-/issues/1736 which says

> I suspect it's because OBS studio uses the right way to transfer the PipeWire buffer contents to CPU memory, which is using an OpenGL context, instead of using mmap() directly. When using gstreamer it'll just mmap() (as well as the begin/end dmabuf read icotls), but that's not a very good method on e.g. amdgpu.

So I guess Chromium shouldn't be doing that either?

grulja:
> This may actually be related to: https://gitlab.gnome.org/GNOME/mutter/-/issues/1736 which says
> 
> > I suspect it's because OBS studio uses the right way to transfer the PipeWire buffer contents to CPU memory, which is using an OpenGL context, instead of using mmap() directly. When using gstreamer it'll just mmap() (as well as the begin/end dmabuf read icotls), but that's not a very good method on e.g. amdgpu.
> 
> So I guess Chromium shouldn't be doing that either?

Indeed. I talked to Jonas Ådahl and we should fix it in WebRTC. I hopefully will be able to look into it soon.

On the other hand he told me that Mutter does it wrong as well, it shouldn't default to use dma-bufs by default when enabled, but it should (and will in future) use it only when the client explicitly requests it, which is something WebRTC doesn't do.

Issue #49, opened 2021-01-07
Use single extension point for extensions, policies and native messaging hosts?


gasinvein:
It's fairly common that a vendor wants to ship several objects of these 3 kinds at once, which may be interdependent. For example, in Fedora the package `fedora-chromium-config` contains an extension and a policy. Things like KeePassXC and Plasma Browser Integration consist of an extension and a native messaging host.
Currently, it's not possible to ship different kinds of such extensions in one bundle. Using single extension point with `merge-dirs` set to `native-messaging-hosts;extensions;policies/managed;policies/recommended` would solve this.

andrunko:
I wouldn't mind to change the extension points to support all dirs or maybe come up with a new extension point for that, but we should avoid changing the existing ones on a backwards incompatible way as that would break existing users which expect the current structure.

Fwiw, there is nothing blocking packages to ship multiple extension points (extension + policy) in the same underlying package (e.g. for unmanaged extensions, as we do on Endless OS)

gasinvein:
Do you know of any other real-world usage of the extension points except EndlessOS?
>there is nothing blocking packages to ship multiple extension points (extension + policy) in the same underlying package

This is possible for unmanaged extensions, but not for native flatpaks. Maintainer has no means to make sure that interdependent extensions are installed simultaneously.

hammerandtongs:
Native messaging host needs a path to a binary, the directory is just for the json metadata around the connection I believe(?). 

Where should that be? Right now ubuntu kde just puts its binary -

/usr/bin/plasma-browser-integration-host

gasinvein:
@hammerandtongs Binary can be stored in the flatpak extension bundle, alongside json metadata. Doing so for Plasma Browser Integration in particular, though, won't be easy due to dependency on Qt/KDE. Also some `flatpak-spawn --host` wrapper may be used.
In any case, this is an interesting question, but out of scope of this issue.

andrunko:
> Do you know of any other real-world usage of the extension points except EndlessOS?
Not that I know of.

> > there is nothing blocking packages to ship multiple extension points (extension + policy) in the same underlying package
> 
> This is possible for unmanaged extensions, but not for native flatpaks. Maintainer has no means to make sure that interdependent extensions are installed simultaneously.

I think the best to do in this case would be to come up with a new extension point then - would be happy to review if someone comes up with a PR but won't have time to look at this atm I'm afraid.

gasinvein:
[...]

Issue #83, opened 2021-03-17
"Open in window" shortcuts not treated as distinct apps by GNOME Shell


wjt:
With Chrome running outside Flatpak, if I go to *☰ → More tools → Create shortcut…* and check "Open as Window", then:

* I get a launcher in my app grid
* The website launches in its own window
* It behaves as an "app": it has its own icon in the GNOME Alt-Tab app switcher and dash (or taskbar on Endless OS)

See this screenshot: I created a MetaFilter shortcut in Chrome, and it has a little dot below it to show that it is running. Other Chrome windows are grouped with Chrome.

![Capture d’écran de 2021-03-17 12-29-59](https://user-images.githubusercontent.com/86760/111481385-cf3a1b00-872a-11eb-9f7d-d5ac2ecc43fb.png)

By contrast, if I do the same thing with this Chromium Flatpak, it's just lumped in with other Chrome windows. I pinned the MetaFilter launcher to the taskbar and clicking it launches the right window, but it's not associated with the MetaFilter launcher.

![Capture d’écran de 2021-03-17 12-30-52](https://user-images.githubusercontent.com/86760/111482966-491ed400-872c-11eb-8ce6-ce62b566316c.png)

I believe this is because GNOME Shell detects sandboxed apps and then requires their `.desktop` file ID to be prefixed with that ID. And in fact this code goes to great lengths to try to support Chrom(e,ium) shortcuts. https://gitlab.gnome.org/GNOME/gnome-shell/blob/ec071ae4c2c747cc432839805d0c3754f647d55a/src/shell-window-tracker.c#L132-215

But of course the `.desktop` files exported by the browser do not have filenames matching the prefix `org.chromium.Chromium.` so all this is defeated.

wjt:
Not sure whether one can just rename the `.desktop` files…

refi64:
Well by complete coincidence someone else had raised this issue and I couldn't figure out why, but it being the app ID prefix makes sense...I'm not aware of any particular reason we can't just rename it, the only trick would be automating the rename in flextop.

Issue #78, opened 2021-03-09
Wrong directories for policies and native-messaging-hosts


iavael:
Extension point definition merges directories with policies to /app/chromium/policies/policies/managed/ and /app/chromium/policies/policies/recommended/
While chromium actually looks for policies at paths /app/chromium/policies/managed/ and /app/chromium/policies/recommended/
```
[pid 730559] openat(AT_FDCWD, "/app/chromium/policies/managed", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = -1 ENOENT (No such file or directory)
[pid 730559] openat(AT_FDCWD, "/app/chromium/policies/recommended", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = -1 ENOENT (No such file or directory)
```

Same goes for native-messaging-hosts: /app/chromium/native-messaging-hosts/native-messaging-hosts/

gasinvein:
Oh, this explains why kerberos auth stopped working for me after some update.

gasinvein:
I guess it was [this](https://github.com/flathub/org.chromium.Chromium/commit/9c892f32b40d870cefcd4a028c0c703db72a5c32#diff-5b52aeb1e04e809e6726675fb35d4aa77ef176390da956f13ab9df250da93aa4L76-R76) change, policies path in the patch changed from `/app/chromium/policies/policies` to `/app/chromium/policies`.

gasinvein:
This should be now fixed by #92

shrublet:
Sorry to necro this, but policies still only load for me at `chromium/policies/policies` with the managed and recommended sub-directories (i.e. `~/.local/share/flatpak/app/org.chromium.Chromium/current/active/files/chromium/policies/policies`). Is this intended? Thanks!

iavael:
@shrublet 
"policies/policies" path is artifact of using org.chromium.Chromium.Policy extension interface, because of this configuration in chromium manifest
```
[Extension org.chromium.Chromium.Policy]
directory=chromium/policies
autodelete=true
no-autodownload=true
subdirectories=true
merge-dirs=policies/managed;policies/recommended
version=1
```
Basically, path fragments from `directory` and from `merge-dirs` are concatenated, and so you get `chromium/policies/policies/{managed,recommended}`

[...]

Issue #53, opened 2021-01-18
Filename is empty in save file dialog when portal enabled


gasinvein:
Subj. When downloading some file with xdg-desktop-portal enabled, the portal save file dialog appears with empty filename.
I'm using portal version 1.8.0 with the gtk backend on Gnome, if that matters.

gasinvein:
Fixed by #92, thanks.

PR #89, opened 2021-04-07
Portal, extension point, application installation fixes

refi64:
Apologies for the delays, I've had most of this like 50-90% ready for a while, but the last two months of the semester have basically been chaos, which delayed this massively.

The main change is that there is now a `org.chromium.Chromium.UnifiedExtension` extension point. You can create the `extensions`, `native-messaging-hosts`, `policies/...` in one extension point, instead of having to split it all up.

I'm looking to merge this in by hand before release, including #86 and #88 with it.

Fixes: #49, #53, #78, #83.

@gasinvein @iavel @wjt please test! This may not build & run perfectly the first time, since as usual, full Flatpak builds on my system take in the order of days, not hours like on Flathub.

### DIFF
--- a/chromium.sh
+++ b/chromium.sh
@@ -9,6 +9,17 @@ get_int32_property() {
     | awk 'match($0, /uint32 ([0-9]+)/, m){print m[1];}'
 }
 
+merge_extensions() {
+  (
+    shopt -s nullglob
+    dest=/app/chromium/extensions/$1
+    mkdir -p $dest
+    for ext in /app/chromium/${1%/*}/$1/*; do
+      ln -s $ext $dest
+    done
+  )
+}
+
 # Check the portal version & make sure it supports expose-pids.
 if [[ $(get_int32_property version) -lt 4 || \
       $(($(get_int32_property supports) & 1)) -eq 0 ]]; then
@@ -24,6 +35,14 @@ if [[ -f "$XDG_CONFIG_HOME/chromium-flags.conf" ]]; then
   unset IFS
 
   set -- "${flags[@]}" "$@"
+fi
+
+if [[ ! -f /app/chromium/extensions/no-mount-stamp ]]; then
+  # Merge all legacy extension points if the symlinks had a tmpfs mounted over
+  # them.
+  merge_extensions native-messaging-hosts
+  merge_extensions policies/managed
+  merge_extensions policies/recommended
 fi
 
 flextop-init
--- a/org.chromium.Chromium.yaml
+++ b/org.chromium.Chromium.yaml
@@ -149,7 +149,7 @@ modules:
     sources:
       - type: git
         url: https://github.com/refi64/flextop
-        commit: 485c332c7ec36a8b74be8f3622fc756ac6263f40
+        commit: 9f66dac84388f520ecab5ac4fa60df6343b7db5b
 
   - name: extensions
     buildsystem: simple
--- a/org.chromium.Chromium.yaml
+++ b/org.chromium.Chromium.yaml
@@ -43,7 +43,7 @@ add-extensions:
   org.chromium.Chromium.Extension:
     version: '1'
     directory: chromium/extensions
-    merge-dirs: extensions
+    merge-dirs: 'extensions;native-messaging-hosts;policies/managed;policies/recommended'
     subdirectories: true
     no-autodownload: true
     autodelete: true
@@ -155,6 +155,8 @@ modules:
     buildsystem: simple
     build-commands:
       - mkdir -p /app/chromium/{extensions,native-messaging-hosts,policies}
+      - for dir in native-messaging-hosts policies; do ln -s /app/chromium/{$dir,extensions}/$dir; done
+      - touch /app/chromium/extensions/no-mount-stamp
 
   - name: pipewire
     buildsystem: meson
--- a/README.md
+++ b/README.md
@@ -3,36 +3,27 @@
 ### Extension points
 
 To avoid having to expose more of the host filesystem in the sandbox but still
-allowing extending Chromium, the following extension points are defined:
-- org.chromium.Chromium.Policy
-- org.chromium.Chromium.Extension
-- org.chromium.Chromium.NativeMessagingHost
+allowing extending Chromium, the `org.chromium.Chromium.Extension` extension
+point is defined.
 
-#### org.chromium.Chromium.Policy
+This extension point is currently on version '1' and will expose any extension
+manifests under the `extensions` subdirectory, policy files under
+`policies/managed` and `policies/recommended`, and [native messaging host
+manifests](https://developer.chrome.com/docs/apps/nativeMessaging/) under
+`native-messaging-hosts`.
 
-This extension point can be used to configure custom Chromium policies and is
-currently on version '1' and will make any policy under the `policies/managed` and
-`policies/recommended` subdirectories available to Chromium.
+#### Legacy extension points
 
-#### org.chromium.Chromium.Extension
-
-Similarly to the above, but for Chromium extensions, this extension point is
-also currently on version '1' and will make any extension under the `extensions`
-subdirectory available to Chromium.
-
-#### org.chromium.Chromium.NativeMessagingHost
-
-Also as above, but for [native messaging host](https://developer.chrome.com/docs/apps/nativeMessaging/)
-support. As the other extension points, this extension point is also currently
-on version '1' and exposes the `native-messaging-hosts` subdirectory to Chromium.
+This application also supports two other extension points:
+`org.chromium.Chromium.Policy` and `org.chromium.Chromium.NativeMessagingHost`.
+These primarily exist for compatibility reasons and should not be used.
 
 #### Using extension points
 
 Extension points can be provided as regular flatpaks and an example is provided
-under `examples/policies/google-safe-search`. Important to note that extension points'
-name must follow the syntax of `<ExtensionPointName>.<id>`, where `<ExtensionPointName>`
-is one of the supported extension points above and `<id>` is a generic id for this
-specific extension point.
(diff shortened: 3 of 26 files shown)


org.chromium.Chromium.yaml
```diff
@@ -32,6 +32,14 @@ add-extensions:
     bundle: true
     autodelete: true
 
+  org.chromium.Chromium.UnifiedExtension:
```
gasinvein, line 35: Hm, wouldn't it be cleaner for the `.Extension` extension point to assume the role of `.UnifiedExtension`, instead of adding a new one?

refi64: My main concern was conceptual, it felt confusing to have `.Policy` and `.NativeMessaingHost` be specific but `.Extension` is "unified".

That being said, the others are essentially "legacy", so maybe this isn't that major. I'll ask around for some other feedback as well.

gasinvein: >it felt confusing to have `.Policy` and `.NativeMessaingHost` be specific but `.Extension` is "unified"

Well, maybe name the new extension point something different altogether to avoid confusion, like `.Addons`?

wjt: In Endless OS we use the current extension points to magic configuration in from the host system, [namely](https://github.com/endlessm/eos-boot-helper/blob/eos3.9/tmpfiles.d/chromium-system-services.conf.in) policies from `/etc/chromium-browser` and extensions from `/usr/share/chromium/extensions`.  (For the curious, we specifically use this to [conditionally enable Cisco OpenDNS Family Shield](https://github.com/endlessm/eos-theme/tree/master/safe-defaults), and [install AdBlock Plus by default](https://github.com/endlessm/eos-browser-tools/blob/master/adblock-plus/cfhdojbkjhnklbpkdaibdccddilifddb.json), respectively. We also have some installations which use custom mandatory policies installed to `/etc/chromium-browser`.)

Because this is done not with 'real' Flatpaks but with configuration that is shipped in the OS itself, we cannot easily update it in lockstep with a new Chromium Flatpak. (We have a mechanism to trigger an app update along with an OS update, but not the other direction.)

So,

> Hm, wouldn't it be cleaner for the `.Extension` extension point to assume the role of `.UnifiedExtension`, instead of adding a new one?

I don't have any strong feelings on the aesthetics of this versus the other suggested names, but, is it possible to do this without breaking Endless OS's existing use of this extension point? Otherwise:

> That being said, the others are essentially "legacy", 

They may be legacy but we really need them to keep working :)

chromium.sh
```diff
@@ -9,6 +9,17 @@ get_int32_property() {
     | awk 'match($0, /uint32 ([0-9]+)/, m){print m[1];}'
 }
 
+merge_extensions() {
+  (
```
wjt, line 13: Why is this subshell needed?

If it's to avoid the 'dest' variable leaking out of this function, you can declare it with `local dest=...` and that will scope it to the function.

chromium.sh
```diff
@@ -9,6 +9,17 @@ get_int32_property() {
...
+    shopt -s nullglob
+    dest=/app/chromium/unified-extensions/$1
+    mkdir -p $dest
+    for ext in /app/chromium/$1/*; do
+      ln -s $ext $dest
+    done
```
wjt, line 19: This doesn't work because `/app` is read-only while the app runs:

```
wjt@camille:~/.local/share$ flatpak run org.chromium.Chromium//test
mkdir: cannot create directory ‘/app/chromium/unified-extensions/extensions’: Read-only file system
ln: failed to create symbolic link '/app/chromium/unified-extensions/extensions': Read-only file system
ln: failed to create symbolic link '/app/chromium/unified-extensions/extensions': Read-only file system
mkdir: cannot create directory ‘/app/chromium/unified-extensions/native-messaging-hosts’: Read-only file system
mkdir: cannot create directory ‘/app/chromium/unified-extensions/policies’: Read-only file system
mkdir: cannot create directory ‘/app/chromium/unified-extensions/policies’: Read-only file system
```

org.chromium.Chromium.yaml
wjt: Or if it's not possible to keep them working, we need to be able to coordinate an OS update for at least best-effort adjustment to the change.

gasinvein: I guess existing extensions will keep working as before if the `.Extension` extension point starts merging not only `extensions` dir but also `policies` and `native-messaging-hosts` dirs. Right?

org.chromium.Chromium.yaml
```diff
@@ -154,7 +162,7 @@ modules:
   - name: extensions
     buildsystem: simple
     build-commands:
-      - mkdir -p /app/chromium/{extensions,native-messaging-hosts,policies}
+      - mkdir -p /app/chromium/{{unified-,}extensions,native-messaging-hosts,policies/{managed,recommended}}
```
wjt, line 165: But now I see the second half of this hunk and I wonder whether the policies bit works in the currently-released version!

wjt: Ah yes, #78, so no, it doesn't. Unfortunately `org.chromium.Chromium.Policy` is not the one being discussed above :-)

org.chromium.Chromium.yaml
```diff
@@ -157,7 +157,7 @@ modules:
     sources:
       - type: git
         url: https://github.com/refi64/flextop
-        commit: 485c332c7ec36a8b74be8f3622fc756ac6263f40
```
wjt, line 152: Nice!

I can't work out how to take a screenshot of Win+Tab but I can work out how to take a screenshot of the Endless OS taskbar.

Capture d’écran de 2021-04-07 21-47-44

chromium.sh
gasinvein: Weird, the extension mount point `/app/chromium/unified-extensions` should be tmpfs for the merge-dirs thing to work. Maybe a tmpfs is mounted over it only when some actual extensions are installed?

chromium.sh
refi64: It's to avoid the `nullglob` from propagating outward. Technically not really something that could be particularly harmful, but IMO with any global state it's better to be safe than sorry.

org.chromium.Chromium.yaml
refi64: > is it possible to do this without breaking Endless OS's existing use of this extension point?

That was pretty much the goal of the symlinks (that apparently aren't working). The old ones are annotated as "legacy" to try to discourage usage (well, they will be when I get around to updating the README I guess), but they'll definitely continue to work.

Also yeah re-using `.Extension` should not be a breaking change (are there even any users of this yet?). For me, it's just the inconsistency of having two extensions based on ways we can extend Chromium (`.Extension` and `.Policy`) but one is actually all-encompassing. Though, I guess, if Endless will probably end up being the only user of `.Policy`, all we'd need is a comment for clarification then?

chromium.sh
refi64: Huh...yeah, I was under the assumption this would be a tmpfs always, but looking at the source, it only mounts the tmpfs when it goes to mount an extension...so it never will if no extension under that point is available. That's definitely a bit of a curveball, I guess I could just throw it up under something guaranteed to be a private tmpfs (`/run/user/1000/app/...` or `/tmp`), it's just a bit uglier that way.

org.chromium.Chromium.yaml
wjt: I suggest reusing `.Extension` and documenting that the others are legacy.

If we're prepared to ignore non-Endless OS users of the other extension points (indeed, per #78 they are broken anyway) then this also dodges the "tmpfs doesn't exist if there aren't any runtimes mounted there" issue. On Endless OS it is guaranteed that `org.chromium.Chromium.Extension.system-extensions=local` always exists.

chromium.sh
gasinvein: An idea: we can create symlinks `/app/chromium/unified-extensions/policies` -> `/app/chromium/policies/policies` at build time, so that the symlinks appear in the sandbox when there are no combined extension mounted, and in the wrapper do a check if `mountpoint /app/chromium/unified-extensions` - then, if tmpfs is mounted over the dir, do the symlink merge.

refi64:
So I went with @gasinvein's quite slick idea (thanks!), hopefully it works... Also threw in a fix for #31 because might as well, it's pretty minor.

org.chromium.Chromium.yaml
```diff
@@ -149,12 +149,14 @@ modules:
...
+        commit: 9f66dac84388f520ecab5ac4fa60df6343b7db5b
 
   - name: extensions
     buildsystem: simple
     build-commands:
       - mkdir -p /app/chromium/{extensions,native-messaging-hosts,policies}
+      - for dir in native-messaging-hosts policies; do ln -s /app/chromium/{$dir,extensions}/$dir; done
```
gasinvein, line 158: If you `mkdir` dirs other than `extensions` prior to creating symlink, you'll end up with symlinks like
`/app/chromium/extensions/policies/policies` -> `/app/chromium/policies`
I guess this is incorrect?
In any case, I suggest using `ln -s -T` to make sure that the symlink either ends up exactly where you told it, or the command fails.

gasinvein: Oh nevermind, I've misread the `{}` expansion.

gasinvein:
Seems like the symlink trick works now. At least the files seems to appear in the right places, unless I've missed something.

gasinvein:
But download via portal seem to be broken now, save file dialog doesn't appear at all, in log
```
Portal returned error: org.freedesktop.portal.Error.InvalidArgument: Expected type 's' for option 'current_name', got 'ay'
```

refi64:
Superceded by #92.

PR #92, opened 2021-04-15
Update Chromium to 90.0.4430.72, combined w/ other open PRs

refi64:
This includes the contents of #86, #88, and #89. @gasinvein this also includes the fix for the terribly stupid portal bug I accidentally added in #89.


refi64:
```
Failed to download sources: module chromium: Can't find file at (null)
```

Well this is certainly interesting...

refi64:
WELL it looks like we hit a compiler bug, @Erick555 looks like we might be switching to LLVM 11 more quickly than I thought 😅 pulled in your commit from #84 so we'll see if that helps...

refi64:
Well, looks like LLVM 11 is resulting in other build errors...

After some thought, I believe the best course of action is going to end up being to move to the upstream toolchain at least on x64, which is where the compile crashes were occurring anyway. This also avoids the original concern of longer build times when using the upstream toolchain (the x64 one is prebuilt).

We could, of course, just update for LLVM 11 and try to switch to that, but the fact that there's *already* something upstream is uses that 11 doesn't is already a pretty harrowing start.

Erick555:
@refi64 please try this patch https://chromium.googlesource.com/chromium/src/+/9909f146b28d56c9c0411329a056ed959b33f76a

Erick555:
I agree about trying upstream toolchain but I'm not sure what will be the fastest fix to unstuck the build. The backlog is quite big now.

Erick555:
Another alternative is to wait for [llvm12](https://github.com/flathub/flathub/pull/2262) on flathub.

refi64:
@Erick555 thanks for the info, I've added a patch to revert that, so hopefully this will build now.

gasinvein:
It works, #53 and #78 seem to be fixed to me.

PR #94, opened 2021-04-16
Install swiftshader subdirectory

GeorgesStavracas:
(Untested)

Spotted when trying to run Chromium with Ozone/Wayland and passing --use-gl=desktop. Chromium fails to load with:

```
Failed to load GLES library: /app/chromium/swiftshader/libGLESv2.so
```

### DIFF
--- a/build-aux/install.sh
+++ b/build-aux/install.sh
@@ -4,7 +4,7 @@ mkdir -p /app/chromium
 
 pushd out/Release
 # Keep file names in sync with build_devel_flatpak.py
-for path in chrome icudtl.dat *.so *.pak *.bin *.png locales MEIPreload; do
+for path in chrome icudtl.dat *.so *.pak *.bin *.png locales MEIPreload swiftshader; do
     cp -rv $path /app/chromium
 done
 popd
--- a/utils/build_devel_flatpak.py
+++ b/utils/build_devel_flatpak.py
@@ -35,6 +35,7 @@ def rewrite_manifest_and_get_env(manifest_file, new_manifest_file, out_dir):
         '*.png',
         'locales',
         'MEIPreload',
+        'swiftshader',
     ]
 
     with open(manifest_file) as fp:


tinywrkb:
https://bugs.chromium.org/p/chromium/issues/detail?id=578890#c380

> me: I don't know if this is a bug but "--use-cmd-decoder=validating --use-gl=desktop" break Ozone Wayland. I used it because of https://bugs.chromium.org/p/chromium/issues/detail?id=1042393 though it seems like I don't need it anymore.

> Igalia's msisov: you don't need to pass gl to Wayland.
> 1) It doesn't support "desktop" - https://source.chromium.org/chromium/chromium/src/+/master:ui/ozone/platform/wayland/gpu/wayland_surface_factory.cc;l=146
> 2) "desktop" on X11 essentially means to use GLX instead of Angle/Glx.

GeorgesStavracas:
Swiftshader would still be necessary, wouldn't it?

tinywrkb:
I would guess so, it's packaged in the Chrome binary/deb release.  
You can reproduce the issue by running like this:
```
flatpak run org.chromium.Chromium --ozone-platform=wayland --enable-features=UseOzonePlatform --disable-gpu
```
It's probably a good idea to test with when Vulkan is disabled an enabled, so add `--enabled-features=Vulkan` and `--disabled-features=Vulkan`.

tinywrkb:
I would guess so, it's packaged in the Chrome binary/deb release.  
You can reproduce the issue by running like this:
```
flatpak run org.chromium.Chromium --ozone-platform=wayland --enable-features=UseOzonePlatform --disable-gpu
```
It's probably a good idea to test with when Vulkan is disabled and enabled, so add `--enabled-features=Vulkan` and `--disabled-features=Vulkan`.

GeorgesStavracas:
Okay. Since the current test build is close to finishing, I'll wait and test it. Then I'll push another update with Vulkan, perhaps in a separate PR?

tinywrkb:
> Okay. Since the current test build is close to finishing, I'll wait and test it. Then I'll push another update with Vulkan, perhaps in a separate PR?

The `--enabled-features=` is a runtime option, no need to rebuild.

The official Chrome binary deb release does not seem to include `libvk_swiftshader.so` so I guess it uses an outdated swiftshader release.  
Running Chrome with `--enabled-features=Vulkan` makes no difference, it doesn't show that Vulkan is enabled in `chrome://gpu`.

GeorgesStavracas:
This is ready for review. The warning about swiftshader is gone.

@tinywrkb  it was pointed out on irc that the correct command is `--enable-features=Vulkan`, and does seem to work even without this branch.

refi64 (approved):

PR #95, opened 2021-04-16
metainfo: Update version to 90.0.4430.72

wjt:
This was missed when updating the tarball.

For completeness I put the release notes URL in here too.


wjt:
Apparently it helps to base your branch on the current top of master.

wjt:
I killed the build but I have verified that the file passes `org.freedesktop.appstream-glib validate`.

wjt:
Merging based on IRC review from @refi64 :)

PR #97, opened 2021-04-19
Use llvm12 extension for build

Erick555:
Also drop compat patches for older llvm versions.


Erick555:
@refi64 so newer llvm works as expected. Perhaps `0023-Revert-base-Rename-SEQUENCE_CHECKER-capability-name` can be dropped too? (I didn't found the reasons it was added).

Side note: does total patch number need to be in subject of every patch? This creates a lot of noise when new patch is removed/added. Also changing patch number thus filename make it harder to follow the patch origin.

refi64:
> Side note: does total patch number need to be in subject of every patch? This creates a lot of noise when new patch is removed/added. Also changing patch number thus filename make it harder to follow the patch origin.

Removing the patch number in the subject is doable..but I'm not sure about the filenames, some patches *do* need to be in order, and having the numbers also helps me be able to just sort through & manage everything more easily.

As for the sequence checker patch, I believe the capability used is only supported in LLVM/Clang Git?

Erick555:
> As for the sequence checker patch, I believe the capability used is only supported in LLVM/Clang Git?

I don't know but I didn't saw distros carrying such patch. I think we should try if dropping it works.

Erick555:
@refi64 dropping the third patch worked. I also splitted patch removal into separate commi and updated total patch number in subjects.

refi64:
Thanks!

PR #106, opened 2021-05-12
Remove Google OAuth2 client ID & secret

### LINKED ISSUE #56: Sync Paused
About once a day, my account gets "paused". Once this happens, my passwords aren't synced any longer and any updates to my other accounts aren't synced.

Even after I log in again, the Chromium Password Manager remains empty and my passwords never resync.

What logs can I provide?

wjt:
As discussed in [this thread][0] and [this wiki page][1], these are used
to sign into Chromium for features like Chrome sync. Unfortunately,
these services are now only available to Google Chrome, not to
third-party builds of Chromium.  Developers can generate their own API
keys and add their Google account to a special mailing list, which will
allow them to sign in to Chromium for development and testing purposes,
but this only grants a very low quota.

Google recommends removing the default client ID & secret in third-party
builds. Do that here.

We keep the API key, which is used to access public APIs such as Safe
Browsing, and is not affected by this policy change.

Fixes #56, as best as is possible.

[0]: https://groups.google.com/a/chromium.org/g/chromium-packagers/c/SG6jnsP4pWM
[1]: https://www.chromium.org/developers/how-tos/api-keys

### DIFF
--- a/build-aux/bootstrap.sh
+++ b/build-aux/bootstrap.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# DO NOT REUSE THE BELOW API KEYS; these are for Flathub only.
+# DO NOT REUSE THE BELOW API KEY; it is for Flathub only.
 # http://lists.debian.org/debian-legal/2013/11/msg00006.html
 tools/gn/bootstrap/bootstrap.py -v --no-clean --gn-gen-args='
     use_sysroot=false
@@ -15,8 +15,6 @@ tools/gn/bootstrap/bootstrap.py -v --no-clean --gn-gen-args='
     clang_use_chrome_plugins=false
     is_official_build=true
     google_api_key="AIzaSyAL6fqCZVFhA7K_qBvz9GO5Z-V1JBcPO0A"
-    google_default_client_id="62778694883-t4a8bf01lumav756cl6ujhhrqsqselq9.apps.googleusercontent.com"
-    google_default_client_secret="eeHsihEqYjhD2zuIROjerl_7"
     treat_warnings_as_errors=false
     proprietary_codecs=true
     ffmpeg_branding="Chrome"


wjt:
So… this deleted my entire Chromium profile. (It at least told me it was going to do so, but did not give any way to cancel!)

ramcq:
Does it behave in the same way if you do this: 
> If this question refers to a chromium derivative, if you set a client_id and secret, you'll have to make sure that the Google signin code is disabled, e.g., by passing the flag --allow-browser-signin=false as the signin integration does not fail gracefully once the quota on the API is exceeded.

(from https://groups.google.com/a/chromium.org/g/chromium-packagers/c/SG6jnsP4pWM/m/c4YTWPixAgAJ )

ramcq:
> I don't actually know whether there is any functionality enabled by the client ID and secret other than the allow-browser-signin, so maybe this suggestion is completely noise/bogus. Also wow that thread is full of very upset people.

OK yeah this is just completely bogus. Merely following the instructions on https://www.groovypost.com/howto/remove-a-google-account-from-chrome/ (disabling the browser sign-in) restarts and creates a new profile for me - and that flag has got to be the build-time version of that toggle.

ramcq:
https://superuser.com/questions/1354910/how-can-i-turn-off-chrome-sync-using-my-g-suite-account-without-losing-my-prof ...? :thinking:

ramcq:
> https://superuser.com/questions/1354910/how-can-i-turn-off-chrome-sync-using-my-g-suite-account-without-losing-my-prof ...?

FWIW I restored my profile from backup and tried this - it worked.

wjt:
Nice discovery!

So, we could pop something into `chromium.sh` to remove those keys from all profiles as a one-time migration step. I think it should be one-time because it is possible to provide environment variables to set the API key, so really determined users can create their own, add themselves to the relevant developer group, `flatpak override --env=…`, and away they go. (Actually, hmm, that's a way I could get my profile back. But I also tried copying my google-chrome profile directory into the right place and that "worked" too.)

wjt:
Capture d’écran de 2021-05-12 22-44-47

This is the error I reproduced by copying my profile from chrome to chromium. I believe it's the same one I saw the first time but wasn't quick enough to screenshot.

refi64:
So I spent a bit more time looking into this and found the following:

The belief that this deletion was *solely* due to this being a managed account. In particular, if the browser is registered with any cloud policies (as with a managed account AFAIK), browser sign-out is [explicitly disabled](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/policy/cloud/user_policy_signin_service.cc;drc=621738fe799111b6206497b0042e369dc4764449;bpv=1;bpt=1;l=217?gsn=ProhibitSignoutIfNeeded&gs=kythe%3A%2F%2Fchromium.googlesource.com%2Fchromium%2Fsrc%3Flang%3Dc%252B%252B%3Fpath%3Dsrc%2Fchrome%2Fbrowser%2Fpolicy%2Fcloud%2Fuser_policy_signin_service.h%23959Kf7UxN943SQoV0cluTlxC4I613MEZLCuNh2WeP7U&gs=kythe%3A%2F%2Fchromium.googlesource.com%2Fchromium%2Fsrc%3Flang%3Dc%252B%252B%3Fpath%3Dsrc%2Fchrome%2Fbrowser%2Fpolicy%2Fcloud%2Fuser_policy_signin_service.cc%23dBruoODvqi_jpVl5ZKBy4wagociBQtYo39cgUI1VHFM). If sign-in / sync is disabled (not clear which of these exactly is tripped by the OAuth creds being removed), then [the code tries to sign out](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/signin/signin_util.cc;drc=621738fe799111b6206497b0042e369dc4764449;bpv=1;bpt=1;l=221?q=IDS_PROFILE_WILL_BE_DELETED_DIALOG_DESCRIPTION&ss=chromium%2Fchromium%2Fsrc&gsn=EnsurePrimaryAccountAllowedForProfile&gs=kythe%3A%2F%2Fchromium.googlesource.com%2Fchromium%2Fsrc%3Flang%3Dc%252B%252B%3Fpath%3Dsrc%2Fchrome%2Fbrowser%2Fsignin%2Fsignin_util.h%2345EAcHQYChVM5FrZhTM_V7gQcQ-tblwXCGmExYR6zHg&gs=kythe%3A%2F%2Fchromium.googlesource.com%2Fchromium%2Fsrc%3Flang%3Dc%252B%252B%3Fpath%3Dsrc%2Fchrome%2Fbrowser%2Fsignin%2Fsignin_util.cc%23vEMvCkoOa5SpwUhBdrQm4tzeqdlUyjf2qVJMfeq4ylg), but will delete the profile if sign-out is not allowed. In addition, I can personally confirm that, for my personal profile, running this PR just signs me out but leaves all the data intact.

In other words, we could bypass this easily...but this feels like a really intentional (security-based?) decision to not preserve the contents of a managed profile if not actively signed in.

wjt:
I was surprised to have my profile obliterated, but at least I knew I was testing this PR. I would be really upset if, one day, I suddenly lost all my bookmarks, saved passwords, etc. without warning, because Chromium auto-updated. Yes, it's a security-based decision but you could argue that this behaviour is not designed to take into account "the entire sync feature was removed from behind the user's back".

I don't suppose there is any way you can tell from the Google API console how many managed accounts have signed into sync, and hence how many users this would affect one way or another.

(I find the "remove two keys from a JSON file, as a one-time migration" idea a lot more appealing than "patching Chromium in perpetuity", FWIW.)

refi64:
@wjt Would it maybe be worth it to first show a disclaimer dialog in that case? From what I can tell, `signout_setting` will be `false` on these profiles, so the migration script could detect that.

wjt:
What problem do we expect a disclaimer dialog to solve? What action would the user take as a result? How would we translate it?

I think we should just remove the keys and then move on.

PR #111, opened 2021-05-19
Remove API keys, unlink profiles, fix WebAssembly OOM errors

### LINKED ISSUE #104: Tab crashes when a WebAssembly OOM occurs
Copied from [this chromium bug report](https://bugs.chromium.org/p/v8/issues/detail?id=11735) (the bug report mentions snapcraft, but I later confirmed the flatpak package has the same problem):

**Version:** Chromium 90
**OS:** Pop!_OS 20.10
**Architecture:** x64

## What steps will reproduce the problem?

1. Install chromium with flatpak, eg
    ``` sh
    flatpak install flathub org.chromium.Chromium
    ```

2. Write the following `main.c`:

    ```c
    #include <stdlib.h>
    #include <stdio.h>

    int main()
    {
        int* p = 3000111000;
        int x = *p;
        printf("val: %d", x);

        return (0);
    }
    ```

3. Compile it and run it using emscripten

    ``` sh
    emcc main.c -o test_oob.html
    emrun --no_browser test_oob.html
    ```

4. Open `http://localhost:6931/test_oob.html` in Chromium

## What is the expected output?

The page logs an exception upon loading, then does nothing.

## What do you see instead?

The page immediately segfaults.

refi64:
- Includes the commit from & supersedes #106
- Unlinks all profiles from Google account data, as mentioned in #106, to avoid profile deletion
- Fixes #104 (...probably, I still haven't managed to build the test file...)

### DIFF
--- a/patches/0022-content-Disable-the-WASM-trap-handler-on-Linux.patch
+++ b/patches/0022-content-Disable-the-WASM-trap-handler-on-Linux.patch
@@ -1,0 +1,26 @@
+From 6bfe197bcf785bd6aaec995b0c2ae605cc6a0a44 Mon Sep 17 00:00:00 2001
+From: Ryan Gonzalez <rymg19@gmail.com>
+Date: Wed, 19 May 2021 16:20:55 -0500
+Subject: [PATCH 22/22] content: Disable the WASM trap handler on Linux
+
+---
+ content/public/common/content_features.cc | 3 +--
+ 1 file changed, 1 insertion(+), 2 deletions(-)
+
+diff --git a/content/public/common/content_features.cc b/content/public/common/content_features.cc
+index cb6a06386a67c..4aed3b34bd146 100644
+--- a/content/public/common/content_features.cc
++++ b/content/public/common/content_features.cc
+@@ -817,8 +817,7 @@ const base::Feature kWebAssemblyThreads {
+ };
+ 
+ // Enable WebAssembly trap handler.
+-#if (defined(OS_LINUX) || defined(OS_CHROMEOS) || defined(OS_WIN) || \
+-     defined(OS_MAC)) &&                                             \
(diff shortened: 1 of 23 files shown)


refi64:
(let's pretend I didn't write the patch, then forget to add it to the manifest)

wjt (approved):
LGTM but I would like someone who is still signed in with a managed account, and has not already accidentally deleted it by testing #106 :-), to give it a spin.

ramcq:
Restored my backup profile (with oauth tokens) and tested, worked fine!

patches/0022-content-Disable-the-WASM-trap-handler-on-Linux.patch
```diff
@@ -0,0 +1,26 @@
+From 6bfe197bcf785bd6aaec995b0c2ae605cc6a0a44 Mon Sep 17 00:00:00 2001
+From: Ryan Gonzalez <rymg19@gmail.com>
+Date: Wed, 19 May 2021 16:20:55 -0500
+Subject: [PATCH 22/22] content: Disable the WASM trap handler on Linux
```
wjt, line 4: Why is this desirable?

refi64: @wjt Slightly worse performance on bounds checks. I'd like to explore an alternative solution (I'd guess the sandbox is blocking the trap handler, but tbh I'm not really sure *why* yet), but this seems like a nasty enough bug (esp if it breaks *all* bounds checks, which I have unfortunately been unable to test) that adding a workaround until we know more is worthwhile.

dsd:
> But I would like someone who is still signed in with a managed account, and has not already accidentally deleted it by testing #106 :-), to give it a spin.

I tested this and it worked here.

PR #117, opened 2021-05-28
Attempt to use the Chromium toolchain for builds

refi64:
Advantages would be:

- No more custom patches to make things build or random compiler crashes (hopefully!).
- We could enable PGO.

The main disadvantage would be increased build times on arm64, where prebuilt binaries aren't available. It would seem the LLVm 11/12 SDK extensions take around 20 minutes to build...which is, on one hand, another 20 minutes on top of a 2 hour build. On the other hand, it's a relatively minor increase for a hopefully decent sized gain.

As of right now, this:

- Will probably not work first try.
- Builds the toolchain on both x64 and arm64, even though x64 has prebuilt binaries available.
- Does not attempt to enable PGO yet. (NOTE: This would not result in a significant build time increase, since the profiles are pre-computed and the files placed in the tarballs.)
- Does not attempt to keep the versions on sync on external-data-checker updates.


refi64:
arm64 took ~32 minutes to build the toolchain before failing during the ffmpeg build (if I had to guess it's trying to use the cross compiler filenames, shouldn't be a hard fix).

Erick555:
@refi64 currently it's sing gcc to build clang/llvm toolchain which is inferior solution for performance. I recommend to use llvm12 ext to build chromium clang/llvm toolchain and revert https://github.com/flathub/org.chromium.Chromium/commit/63f22bbede2760369ffcd9253d7e52ac65fc89c9 along the way.

refi64:
@Erick555 thanks for the heads up, it's a bit trickier than that but I think it's workable

Erick555:
@refi64 you forgot toy remove [.  /usr/lib/sdk/llvm11/enable.sh](https://github.com/flathub/org.chromium.Chromium/blob/chromium-toolchain/build-aux/build.sh#L5) from build.sh which causes failure for aarch64.

The total build time for x86 is 2.5h which on par with llvm12 extension build before.

Great work.

refi64:
@Erick555 indeed, this may be the most painful forget-to-update mistake I have made in quite some time :sweat_smile:

refi64:
Nice upside: with PGO enabled, our speedometer benchmarks go from ~67 to ~81 (at least on my system).

Erick555:
x86 build time with pgo is 1:36:47. Looks like bundled compiler is insanely optimized.

Minor nitpick: for aarch64 internal clang is linked with gnu ld while forcing it to lld may yield another (perhaps not that significant) optimization. I'm not sure if this is worth patching build script again though.

Erick555:
@refi64 is there something more to do here? Perf improvements sound quite significant to release it.

refi64:
@Erick555 was hoping to get some arm64 testers...but this should probably still work, just gonna cross my fingers on this one!

PR #123, opened 2021-06-21
Fix NETLINK bind errors

### LINKED ISSUE #50: no proxy extension is working
tried about 5 of proxy/vpn extensions for Chrome from official store, those which worked with ubuntu ppa/snap versions are refusing to change external IP. something is seriously weird with networking in flatpak build.

refi64:
*Maybe* fixes #50, based on the comments from inside, not entirely sure.


refi64:
Would appreciate if someone else could test, but some basic experimentation is showing that this does *not* in fact fix the proxy extension issue.

refi64:
Fixed upstream, and this doesn't seem to cause any problems to justify taking on a new patch, so closing.

PR #126, opened 2021-07-01
Re-enable XDG file chooser portal

### LINKED ISSUE #33: GTK3 File Picker on KDE Plasma
KDE Plasma users naturally prefer native file-picker. Google Chrome &  Chromium packaged by distrubutions both use KDialog (nice middle-ground for now).
Snap and Flatpak versions of Chromium use GTK3 File Picker which comparing to KDialog lacks usabitility feautures & brings interface Inconsistency.

vchernin:
Reverses part of cba74b8f9f246cc144231e06f4266a1b05d79757, since https://github.com/flatpak/xdg-desktop-portal/issues/545 is now fixed.

Fixes #33 

The freedesktop runtime appears to be on a [git snapshot of XDG desktop portal 1.8.1](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/master/elements/components/xdg-desktop-portal-base.bst), so this should work.

### DIFF
--- a/patches/0018-Add-support-for-the-XDG-file-chooser-portal.patch
+++ b/patches/0018-Add-support-for-the-XDG-file-chooser-portal.patch
@@ -91,7 +91,7 @@ index 29a222be4493a..0e7e312b8eee1 100644
 +#if defined(OS_LINUX)
 +// Enables use of the XDG file chooser portal for file selection.
 +const base::Feature kXdgFileChooserPortal{"XdgFileChooserPortal",
-+                                          base::FEATURE_DISABLED_BY_DEFAULT};
++                                          base::FEATURE_ENABLED_BY_DEFAULT};
 +#endif  // defined(OS_LINUX)
 +
 +#if defined(OS_MACOS)


Erick555:
I think it's the host xdg portal version that matters, not runtime.

vchernin:
[Fedora is on 1.8.1](https://src.fedoraproject.org/rpms/xdg-desktop-portal), so hopefully that should work. I'll test this out in an Ubuntu 20.04 VM as well.

vchernin:
On each distro I installed the test build and disabled filesystem=home file system access in flatseal. I tried uploading both files and folders on websites with a file uploader button.

- Fedora 34 works and correctly uses the xdg file chooser (host has portal 1.8.1)
- Ubuntu 20.04 refuses to install due to the flatpak version being old. The same happens with the stable build. I assume the same will be true for Debian 10 due to being from 2 years ago.
- Ubuntu 21.04 works and correctly uses the xdg file chooser (host has portal 1.8.1)
- Pop os 20.10 works and correctly uses the xdg file chooser ([host has 1.8.0-1](https://packages.ubuntu.com/groovy/xdg-desktop-portal)). It seems that the portal works correctly despite the host portal being older.

Still hopefully we can get more tests on distros with even older host portal versions.

vchernin:
It appears xdg desktop portal 1.8.0-1 in pop 20.10 [did not have the fix, Debian's 1.8.0-1 was made in Septemeber 2020](https://metadata.ftp-master.debian.org/changelogs//main/x/xdg-desktop-portal/xdg-desktop-portal_1.8.1-1_changelog). I assume ubuntu and pop os used the same package. 

The fix couldn't have been there [since it was only finished in Dec 2020](https://github.com/flatpak/xdg-desktop-portal/issues/545). This seems to indicate that having xdg desktop portal 1.8.1 in the runtime is enough to get everything working, and having 1.8.0(-1) in the host isn't an issue.

refi64:
This is...odd. To be clear, the file chooser portal worked, but then trying to open the resulting files would fail (don't remember if it was the "show in folder" option or the link itself). Regardless, I wonder if there's anyone we can ping to ask to push out 1.8.1 to Ubuntu, seems weird they're holding back on a bugfix release...

It's worth noting that a newer version of this patch *has* been merged into upstream Chromium for version 93(?), so I'm not sure how worthwhile it is to enable now vs just waiting on the (slightly improved) version in upstream.

vchernin:
I tested Ubuntu 20.10 as well, it also has 1.8.0-1. I noticed that the "open in folder" button indeed doesn't work.

However 20.10 will be EOL on July 22, so I think it's unlikely anyone would publish a fix at this point. 21.04 has 1.8.1 and "open in folder" works.

[It found the new patch version you mentioned](https://bugs.chromium.org/p/chromium/issues/detail?id=885292). Assuming Chromium doesn't revert the change it should probably make it for [Chromium 93](https://www.chromestatus.com/features/schedule) on August 31st. Seeing as the new patch should not use the portal if it detects a problem, maybe it makes some sense to try backport it here as it would help test it? Does Chromium Flatpak backport such a change? (I'm not even sure if that new patch would build against 91 correctly).

One thing I also noticed is that drag and drop between Nautilus and Chromium doesn't work without granting file system access, but I'm pretty sure this is since Nautilus isn't on GTK4 yet, and there isn't anything we can really do here. [See this issue](https://github.com/Rafostar/clapper/issues/75).

I was about to send my comment but then I ran into another bug... With this build try to download a file and then save it somwhere like downloads. However, when you click open in folder it will indeed open the file manager, but in something like `/run/user/1000/doc/8584b373`. The file is actually correctly saved in Downloads as well, but the user is directed to the wrong directory. This on Fedora 34 and Ubuntu 21.04 with the up to date portal versions. It seems [that this is solved in master](https://github.com/flatpak/xdg-desktop-portal/pull/552) but it probably won't be in stable distros for a while...

Erick555:
> I wonder if there's anyone we can ping to ask to push out 1.8.1 to Ubuntu, seems weird they're holding back on a bugfix release

This is rather ubuntu/debian official policy for holding back any updates unless there are convincing reasons to do them. I think ubuntu may just use what debian unstable has at time of release. LTS version has still 1.6 and even flatpak official ppa is on 1.7.x so the problem is deeper.

refi64:
This should be enabled upstream soon with a slightly cleaner patch that has some bug fixes, at this point I'm just going to wait for that to come (and try and resolve these portal issues by then).

PR #130, opened 2021-07-21
Update to Chromium 92 w/ massive patch refactor

refi64:
NOTE: A full build has not yet been run locally, so this may still fail.

This updates all patches for Chromium 92 but also completely reworks the
way the patches are created and updated, as well as reducing the patch
count itself. In particular:

- An external tool (currently private but will be public shortly) is
  used to manage the workflow of patch updates, including generating
  the FFmpeg configuration and build shells. As a result, the current
  tools in utils/ have been removed (they used out-of-date mechanisms
  anyway that are no longer needed).
- Udev is now enabled. Disabling it was a huge mistake on my part, as I
  was under the impression that it did not work *at all*, whereas in
  practice, devices can at least be listed (monitoring is what does not
  work). With this, we should be able to drop the patches related to
  fixing the WebUSB and WebHID services on no-udev builds.
- The patches for using system-wide Node and OpenJDK have been removed,
  as the build scripts now just use symlinks to "fake" the existence of
  the Chromium-local Node/OpenJDK using the system versions.
- The patch for arm64 tool prefixes was removed, as now a symlink is
  used to create a tool with the proper prefix regardless.
- The Chromecast patch was removed, as the media router component is no
  longer actually used for Chromecast, thus it works out-of-the-box on
  Chromium builds nowadays.
- The generated FFmpeg configuration itself is in a separate patch,
  which should make regenerating it *significantly* easier now.


refi64:
Merged by hand.

Issue #121, opened 2021-06-18
Pass PipeWire capturer flag by default


vchernin:
PipeWire 0.3 currently [is built by default](https://github.com/flathub/org.chromium.Chromium/pull/88). However, for screen sharing to work on Wayland users need to run with `--enable-features=WebRTCPipeWireCapturer` or enable it in [chrome://flags](chrome://flags/#enable-webrtc-pipewire-capturer). I suggest passing the flag by default [in the .desktop file](https://github.com/flathub/com.slack.Slack/pull/118/commits/8552181dc907f00b9682c784c8876205098d3617), or [by building with it](https://github.com/jitsi/jitsi-meet-electron/pull/584/commits/0e0483cbc52a9cad1fef51ed5abb846bd6445b11) (if that’s feasible to do).

[Jitsi Meet Electron](https://github.com/jitsi/jitsi-meet-electron/pull/584) and [Slack Flatpak](https://github.com/flathub/com.slack.Slack/pull/118/) both pass this flag by default. [Slack support even suggests Wayland users pass the flag with Slack](https://github.com/flathub/com.slack.Slack/issues/101#issuecomment-808430530).

Of course this flag is there since Chromium’s PipeWire implementation is not yet perfect. There are bugs [like the dual pop up “dialog hell”](https://jgrulich.cz/2020/12/18/webrtc-chromium-updates-in-2020/) that can occur. However, I personally think mostly working screen sharing on Wayland by default is better than no screen sharing at all. Especially since it’s not obvious  users need to enable a flag to get working screen sharing. There is also a benefit of getting wider testing of this feature and hopefully useful PipeWire/Chromium bug reports. Also, running with the flag doesn’t seem to regress for X11 users.

If this isn’t a sensible idea perhaps a mention of the flag could be added to the Flathub page description. Also tab sharing works fine without the flag as that doesn’t use PipeWire.

Note there are currently WebRTC bugs [with PipeWire screen sharing on AMD](https://github.com/flathub/org.chromium.Chromium/pull/88#issuecomment-823788246), while Intel should work properly (it does for me). I don’t believe adding this flag would regress on AMD (as it doesn’t seem to work currently anyway), but it’s probably worth testing.

wjt:
Sounds like a good idea in principle if it does not regress X11 usage. Perhaps it could be set in `chromium.sh`, the entry point in this package which invokes the browser. Fancy writing a PR?

vchernin:
[...]

PR #131, opened 2021-07-21
Enable PipeWire capturer by default

vchernin:
Should only improve Wayland experience and not regress X11.
See https://github.com/flathub/org.chromium.Chromium/issues/121

If there's a cleaner/better way of doing this I can try and do so.

### DIFF
--- a/chromium.sh
+++ b/chromium.sh
@@ -40,6 +40,9 @@ fi
 # Merge the flags.
 if [[ -f "$XDG_CONFIG_HOME/chromium-flags.conf" ]]; then
   IFS=$'\n'
+  # Append non-default PipeWire flag for Wayland screen sharing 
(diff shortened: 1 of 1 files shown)


refi64:
The problem with modifying chromium-flags.conf like this is that enable/disable-features arguments aren't cumulative, i.e. if you do `--enable-features=X` and then later `--enable-features=Y`, only `Y` will be enabled, not both. (This *is* a problem the Chrome Flatpak has by passing --disable-features, but that argument is used more rarely IME...)

I'm working on a centralized system for Chrom-e/-ium Flatpaks to read the flag files & pass arguments and such, and that'll include a system for enabling features without stomping over other feature flags and such. That being said, due to the above issue, I'll close this PR for now (but I'll make sure to post in & close the relevant issue when this is updated).

PR #134, opened 2021-07-25
Use Cobalt to start Chromium

refi64:


refi64:
Works for me, but I'll leave this for a few days with Chrome/Edge/Brave betas all using it to make sure it's ready for our stable branch here.

refi64:
Merged into #151.

PR #135, opened 2021-07-26
[WIP] WebRTC: Add proper DMA-BUFs support

grulja:
This is just a preliminary patch, I haven't submitted it to upstream for review yet. It would be helpful if I can get some feedback whether it works or not.

### DIFF
--- a/patches/chromium/PipeWire-capturer-implement-proper-DMA-BUFs-support.patch
+++ b/patches/chromium/PipeWire-capturer-implement-proper-DMA-BUFs-support.patch
@@ -1,0 +1,966 @@
+From 9d0df4e7739487f01e8fbaa4b519bfbc1b661f52 Mon Sep 17 00:00:00 2001
+From: Jan Grulich <grulja@gmail.com>
+Date: Tue, 27 Jul 2021 10:35:28 +0200
+Subject: [PATCH] PipeWire capturer: implement proper DMA-BUFs support
+
+Currently both KWin (KDE) and Mutter (GNOME) window managers don't
+use DMA-BUFs by default, but only when client asks specifically for
+them (KWin) or when experimental DMA-BUF support is enabled (Mutter).
+While current implementation works just fine on integrated graphics
+cards, it causes issues on dedicated GPUs (AMD and NVidia) where the
+code either crashes or screensharing is slow and unusable.
+
+To fix this, DMA-BUFs has to be opened using OpenGL context and not
+being directly mmaped(). This implementation requires to use DMA-BUF
+modifiers, as they are now mandatory for DMA-BUFs usage.
+
+Documentation for this behavior can be found here:
+https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/doc/dma-buf.dox
+---
+ modules/desktop_capture/BUILD.gn              |  23 +
+ .../linux/base_capturer_pipewire.cc           | 217 +++++++--
+ .../linux/base_capturer_pipewire.h            |   8 +
+ modules/desktop_capture/linux/egl_dmabuf.cc   | 420 ++++++++++++++++++
+ modules/desktop_capture/linux/egl_dmabuf.h    |  58 +++
+ modules/desktop_capture/linux/pipewire03.sigs |   1 +
+ 6 files changed, 682 insertions(+), 45 deletions(-)
+ create mode 100644 modules/desktop_capture/linux/egl_dmabuf.cc
+ create mode 100644 modules/desktop_capture/linux/egl_dmabuf.h
+
+diff --git a/third_party/webrtc/modules/desktop_capture/BUILD.gn b/third_party/webrtc/modules/desktop_capture/BUILD.gn
+index 25b92bed45..9562c168eb 100644
+--- a/third_party/webrtc/modules/desktop_capture/BUILD.gn
++++ b/third_party/webrtc/modules/desktop_capture/BUILD.gn
+@@ -230,6 +230,18 @@ if (is_linux || is_chromeos) {
+           ignore_libs = true
+         }
+       }
++      pkg_config("gbm") {
++        packages = [ "gbm" ]
++      }
++      pkg_config("egl") {
++        packages = [ "egl" ]
++      }
++      pkg_config("epoxy") {
++        packages = [ "epoxy" ]
++      }
++      pkg_config("libdrm") {
++        packages = [ "libdrm" ]
++      }
+     } else {
+       pkg_config("pipewire") {
+         packages = [ "libpipewire-0.2" ]
+@@ -562,10 +574,21 @@ rtc_library("desktop_capture_generic") {
+       "linux/base_capturer_pipewire.h",
+     ]
+ 
++    if (rtc_pipewire_version == "0.3") {
++      sources += [
++        "linux/egl_dmabuf.cc",
++        "linux/egl_dmabuf.h",
++      ]
++    }
++
+     configs += [
+       ":pipewire_config",
+       ":gio",
+       ":pipewire",
++      ":gbm",
++      ":egl",
++      ":epoxy",
++      ":libdrm",
+     ]
+ 
+     if (!rtc_link_pipewire) {
+diff --git a/third_party/webrtc/modules/desktop_capture/linux/base_capturer_pipewire.cc b/third_party/webrtc/modules/desktop_capture/linux/base_capturer_pipewire.cc
+index e5d001e476..3be118f928 100644
+--- a/third_party/webrtc/modules/desktop_capture/linux/base_capturer_pipewire.cc
++++ b/third_party/webrtc/modules/desktop_capture/linux/base_capturer_pipewire.cc
+@@ -22,8 +22,10 @@
+ #include <sys/ioctl.h>
+ #include <sys/mman.h>
+ #include <sys/syscall.h>
++#include <unistd.h>
+ 
+ #include <memory>
++#include <string>
+ #include <utility>
+ 
+ #include "absl/memory/memory.h"
+@@ -94,6 +96,78 @@ static void SyncDmaBuf(int fd, uint64_t start_or_end) {
+   }
+ }
+ 
++#if !PW_CHECK_VERSION(0, 3, 29)
++#define SPA_POD_PROP_FLAG_MANDATORY (1u << 3)
++#endif
++#if !PW_CHECK_VERSION(0, 3, 33)
++#define SPA_POD_PROP_FLAG_DONT_FIXATE (1u << 4)
++#endif
++
++struct pw_version {
++  int major;
++  int minor;
++  int micro;
++};
++
++static pw_version ParsePipeWireVersion(const char* version) {
++  pw_version pw_version;
++  sscanf(version, "%d.%d.%d", &pw_version.major, &pw_version.minor,
++         &pw_version.micro);
++  return pw_version;
++}
++
++#if PW_CHECK_VERSION(0, 3, 0)
++static spa_pod* BuildFormat(
++    spa_pod_builder* builder,
++    uint32_t format,
++    const std::vector<uint64_t>& modifiers = std::vector<uint64_t>()) {
++  int c = 0;
++  spa_pod_frame f[2];
++  spa_rectangle pw_min_screen_bounds = spa_rectangle{1, 1};
++  spa_rectangle pw_max_screen_bounds = spa_rectangle{UINT32_MAX, UINT32_MAX};
++
++  spa_pod_builder_push_object(builder, &f[0], SPA_TYPE_OBJECT_Format,
++                              SPA_PARAM_EnumFormat);
++  spa_pod_builder_add(builder, SPA_FORMAT_mediaType,
++                      SPA_POD_Id(SPA_MEDIA_TYPE_video), 0);
++  spa_pod_builder_add(builder, SPA_FORMAT_mediaSubtype,
++                      SPA_POD_Id(SPA_MEDIA_SUBTYPE_raw), 0);
++  spa_pod_builder_add(builder, SPA_FORMAT_VIDEO_format, SPA_POD_Id(format), 0);
++
++  if (modifiers.size()) {
++    pw_version pw_version = ParsePipeWireVersion(pw_get_library_version());
++
++    // SPA_POD_PROP_FLAG_DONT_FIXATE can be used with PipeWire >= 0.3.33
++    if (pw_version.major >= 0 && pw_version.minor >= 3 &&
++        pw_version.micro >= 33) {
++      spa_pod_builder_prop(
++          builder, SPA_FORMAT_VIDEO_modifier,
++          SPA_POD_PROP_FLAG_MANDATORY | SPA_POD_PROP_FLAG_DONT_FIXATE);
++    } else {
++      spa_pod_builder_prop(builder, SPA_FORMAT_VIDEO_modifier,
++                           SPA_POD_PROP_FLAG_MANDATORY);
++    }
++    spa_pod_builder_push_choice(builder, &f[1], SPA_CHOICE_Enum, 0);
++    // mofifiers from the array
++    for (auto it = modifiers.begin(); it != modifiers.end(); it++) {
++      spa_pod_builder_long(builder, *it);
++      if (c++ == 0) {
++        spa_pod_builder_long(builder, *it);
++      }
++    }
++    spa_pod_builder_pop(builder, &f[1]);
++  }
++
++  spa_pod_builder_add(
++      builder, SPA_FORMAT_VIDEO_size,
++      SPA_POD_CHOICE_RANGE_Rectangle(
++          &pw_min_screen_bounds, &pw_min_screen_bounds, &pw_max_screen_bounds),
++      0);
++
++  return static_cast<spa_pod*>(spa_pod_builder_pop(builder, &f[0]));
++}
++#endif
++
+ class ScopedBuf {
+  public:
+   ScopedBuf() {}
+@@ -120,9 +194,9 @@ class ScopedBuf {
+   unsigned char* get() { return map_; }
+ 
+  protected:
+-  unsigned char* map_ = nullptr;
++  unsigned char* map_ = static_cast<unsigned char*>(MAP_FAILED);
+   int map_size_;
+-  bool is_dma_buf_;
++  bool is_dma_buf_ = false;
+   int fd_;
+ };
+ 
+@@ -313,6 +387,9 @@ void BaseCapturerPipeWire::OnStreamFormatChanged(void* data,
+   auto size = height * stride;
+ 
+   that->desktop_size_ = DesktopSize(width, height);
++#if PW_CHECK_VERSION(0, 3, 0)
++  that->modifier_ = that->spa_video_format_.modifier;
++#endif
+ 
+   uint8_t buffer[1024] = {};
+   auto builder = spa_pod_builder{buffer, sizeof(buffer)};
+@@ -320,11 +397,18 @@ void BaseCapturerPipeWire::OnStreamFormatChanged(void* data,
+   // Setup buffers and meta header for new format.
+   const struct spa_pod* params[3];
+ #if PW_CHECK_VERSION(0, 3, 0)
++  const auto buffer_types =
++      spa_pod_find_prop(format, nullptr, SPA_FORMAT_VIDEO_modifier)
++          ? (1 << SPA_DATA_DmaBuf) | (1 << SPA_DATA_MemFd) |
++                (1 << SPA_DATA_MemPtr)
++          : (1 << SPA_DATA_MemFd) | (1 << SPA_DATA_MemPtr);
++
+   params[0] = reinterpret_cast<spa_pod*>(spa_pod_builder_add_object(
+       &builder, SPA_TYPE_OBJECT_ParamBuffers, SPA_PARAM_Buffers,
+       SPA_PARAM_BUFFERS_size, SPA_POD_Int(size), SPA_PARAM_BUFFERS_stride,
+       SPA_POD_Int(stride), SPA_PARAM_BUFFERS_buffers,
+-      SPA_POD_CHOICE_RANGE_Int(8, 1, 32)));
++      SPA_POD_CHOICE_RANGE_Int(8, 1, 32), SPA_PARAM_BUFFERS_dataType,
++      SPA_POD_CHOICE_FLAGS_Int(buffer_types)));
+   params[1] = reinterpret_cast<spa_pod*>(spa_pod_builder_add_object(
+       &builder, SPA_TYPE_OBJECT_ParamMeta, SPA_PARAM_Meta, SPA_PARAM_META_type,
+       SPA_POD_Id(SPA_META_Header), SPA_PARAM_META_size,
+@@ -499,6 +583,12 @@ BaseCapturerPipeWire::~BaseCapturerPipeWire() {
+   }
+ }
+ 
++#if PW_CHECK_VERSION(0, 3, 0)
++void BaseCapturerPipeWire::InitEGL() {
++  egl_dmabuf_ = std::make_unique<EglDmaBuf>();
++}
++#endif
++
+ void BaseCapturerPipeWire::InitPortal() {
+   cancellable_ = g_cancellable_new();
+   g_dbus_proxy_new_for_bus(
+@@ -619,8 +709,6 @@ pw_stream* BaseCapturerPipeWire::CreateReceivingStream() {
+     return nullptr;
+   }
+ #endif
+-  spa_rectangle pwMinScreenBounds = spa_rectangle{1, 1};
+-  spa_rectangle pwMaxScreenBounds = spa_rectangle{UINT32_MAX, UINT32_MAX};
+ 
+   pw_properties* reuseProps =
+       pw_properties_new_string("pipewire.client.reuse=1");
+@@ -630,24 +718,45 @@ pw_stream* BaseCapturerPipeWire::CreateReceivingStream() {
+   auto stream = pw_stream_new(pw_remote_, "webrtc-consume-stream", reuseProps);
+ #endif
+ 
++#if PW_CHECK_VERSION(0, 3, 0)
++  const spa_pod** params;
++  spa_pod_builder builder;
++  uint32_t n_formats, n_params = 0;
++  uint8_t buffer[2048] = {};
++  std::vector<uint64_t> modifiers;
++
++  builder = spa_pod_builder{buffer, sizeof(buffer)};
++  uint32_t formats[] = {
++      SPA_VIDEO_FORMAT_BGRA,
++      SPA_VIDEO_FORMAT_RGBA,
++      SPA_VIDEO_FORMAT_BGRx,
++      SPA_VIDEO_FORMAT_RGBx,
++  };
++
++  n_formats = sizeof(formats) / sizeof(formats[0]);
++  params = static_cast<const spa_pod**>(
++      calloc(2 * n_formats, sizeof(struct spa_param*)));
++
++  for (uint32_t i = 0; i < n_formats; i++) {
++    pw_version pw_version = ParsePipeWireVersion(pw_get_library_version());
++
++    // Modifiers can be used with PipeWire >= 0.3.29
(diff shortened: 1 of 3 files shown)


grulja:
bot, build

refi64:
(Killed the first of the two since I believe they were dups) Not needed right now, but before merge, could this be adjusted for the patch to include the commit headers (i.e. use `git format-patch` instead of `git diff`)? It makes resolving conflicts infinitely easier on my end.

refi64:
Right, also I believe I know someone who this was crashing for previously, so I'll ask them to test this once it's built.

vchernin:
I tested this out on a recent amd ryzen 4000 laptop (exclusively amd cpu+gpu) with a Fedora 34 live usb.
 
1. Install this build
2. Enable pipewire in chrome://flags and restart chromium.
3. Try to screen share windows and the entire screen with: https://mozilla.github.io/webrtc-landing/gum_test.html

Strangely both the current stable and this test build seem to work without a problem. Does anyone know a specific case or website where the issue occurs, or what I should be looking for? Maybe it works differently on a live usb?

grulja:
> I tested this out on a recent amd ryzen 4000 laptop (exclusively amd cpu+igpu) with a Fedora 34 workstation live usb.
> 
> 1. Install this build
> 2. Enable pipewire in chrome://flags and restart chromium.
> 3. Try to screen share windows and the entire screen with: https://mozilla.github.io/webrtc-landing/gum_test.html
> 
> Strangely both the current stable and this test build seem to work without a problem. Does anyone know a specific case or website where the issue occurs, or what I should be looking for? Maybe it works differently on a live usb? Or maybe the bug isn't about AMD always not working, but dedicated gpus behaving differently than integrated ones?

Did you enable experimental DMA-BUF support in GNOME? 

I think `gsettings set org.gnome.mutter experimental-features "['dma-buf-screen-sharing']"` does the trick.

vchernin:
The command doesn't seem to change any behaviour unfortunately. I am able to succesfully screen share on stable and on this test build. 

Still not sure whether it's due to the live usb or hardware or something else. Maybe I'll install Fedora on this device later to see for sure.

grulja:
> (Killed the first of the two since I believe they were dups) Not needed right now, but before merge, could this be adjusted for the patch to include the commit headers (i.e. use `git format-patch` instead of `git diff`)? It makes resolving conflicts infinitely easier on my end.

Will do once I submit this for review. I will do some polishing first. This was just first iteration and something people can try and give me feedback.

vchernin:
Some logs:
[pw-mon.log](https://github.com/flathub/org.chromium.Chromium/files/6883700/pw-mon.log)

[d-bus monitor.log](https://github.com/flathub/org.chromium.Chromium/files/6883699/d-bus.monitor.log)

Both were while screen sharing with Chromium stable on a fedora 34 live usb with the dma-buf experimental option enabled. Still can't reproduce any crash or anything, I'll see if I can do an install later.

grulja:
Here is link to the MR if anyone is interested: https://webrtc-review.googlesource.com/c/src/+/227022/

patches/chromium/PipeWire-capturer-implement-proper-DMA-BUFs-support.patch
```diff
@@ -0,0 +1,421 @@
...
++    }
++
++    gbm_import_fd_data importInfo = {
++        static_cast<int>(spaBuffer->datas->fd),
++        static_cast<uint32_t>(desktop_size_.width()),
++        static_cast<uint32_t>(desktop_size_.height()),
++        static_cast<uint32_t>(spaBuffer->datas[0].chunk->stride),
++        GBM_BO_FORMAT_ARGB8888};
++    gbm_bo* imported = gbm_bo_import(gbm_device_, GBM_BO_IMPORT_FD, &importInfo,
++                                     GBM_BO_USE_SCANOUT);
```
emersion, line 249: No need for `GBM_BO_USE_SCANOUT` here

patches/chromium/PipeWire-capturer-implement-proper-DMA-BUFs-support.patch
```diff
@@ -0,0 +1,421 @@
...
++      RTC_LOG(LS_ERROR) << "Failed to process DMA buffer.";
++      return;
++    }
++
++    gbm_import_fd_data importInfo = {
++        static_cast<int>(spaBuffer->datas->fd),
++        static_cast<uint32_t>(desktop_size_.width()),
++        static_cast<uint32_t>(desktop_size_.height()),
++        static_cast<uint32_t>(spaBuffer->datas[0].chunk->stride),
++        GBM_BO_FORMAT_ARGB8888};
```
emersion, line 247: Are we sure the buffer will have this format?

patches/chromium/PipeWire-capturer-implement-proper-DMA-BUFs-support.patch
```diff
@@ -0,0 +1,421 @@
...
+   explicit BaseCapturerPipeWire(CaptureSourceType source_type);
+   ~BaseCapturerPipeWire() override;
+ 
+@@ -81,6 +91,12 @@ class BaseCapturerPipeWire : public DesktopCapturer {
+   pw_stream_events pw_stream_events_ = {};
+ 
+   struct spa_video_info_raw spa_video_format_;
++
++  bool egl_initialized_ = false;
++  gint32 drm_fd_ = 0;                 // for GBM buffer mmap
```
emersion, line 404: This should default to -1. 0 is a valid FD number.

emersion:
Hm, maybe it would be more appropriate to leave comments on Gerrit?

patches/chromium/PipeWire-capturer-implement-proper-DMA-BUFs-support.patch
```diff
@@ -0,0 +1,421 @@
...
+   params[1] = reinterpret_cast<spa_pod*>(spa_pod_builder_add_object(
+       &builder, SPA_TYPE_OBJECT_ParamMeta, SPA_PARAM_Meta, SPA_PARAM_META_type,
+       SPA_POD_Id(SPA_META_Header), SPA_PARAM_META_size,
+@@ -499,6 +529,86 @@ BaseCapturerPipeWire::~BaseCapturerPipeWire() {
+   }
+ }
+ 
++void BaseCapturerPipeWire::InitEGL() {
++  // Initialize EGL for DMA-BUF support
++  drm_fd_ = open("/dev/dri/renderD128", O_RDWR);
```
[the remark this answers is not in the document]
grulja: I already implemented that and updated MR in upstream, but not here.

columbarius:
Is there a Repo where this patch is applied? It would be nice to make to announcement of capabilities via pipewire compatible with https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1939.

grulja:
> Hm, maybe it would be more appropriate to leave comments on Gerrit?

Yes please: https://webrtc-review.googlesource.com/c/src/+/227022/

grulja:
> Is there a Repo where this patch is applied? It would be nice to make the announcement of capabilities via pipewire compatible with https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1939.

I don't think it matters that much, if we announce type which is not supported by Mutter, but supported by KWin for example, it should negotiate a format that is supported by both sides.

patches/chromium/PipeWire-capturer-implement-proper-DMA-BUFs-support.patch
grulja: It's either GBM_BO_FORMAT_ARGB8888 or GBM_BO_FORMAT_XRGB8888. I don't think it matters, either the alpha channel is ignored or not, it will be processed into correct format later anyway.

emersion: Where is this ensured?

refi64:
Thanks for this! I left some comments on the Gerrit patch set.

columbarius:
> I don't think it matters that much, if we announce type which is not supported by Mutter, but supported by KWin for example, it should negotiate a format that is supported by both sides.


It matters, because after that patch it is required via the format announcement to announce a modifier. Else announcing only dataType SPA_DATA_DmaBuf without a modifier before will be interpreted as a client only able to mmap dmabufs, which will mostly be rejected, and mutter will fall back to shm. This behaviour was discussed at guadec last week and hasn't made it into documentation.

grulja:
> > I don't think it matters that much, if we announce type which is not supported by Mutter, but supported by KWin for example, it should negotiate a format that is supported by both sides.
> 
> It matters, because after that patch it is required via the format announcement to announce a modifier. Else announcing only dataType SPA_DATA_DmaBuf without a modifier before will be interpreted as a client only able to mmap dmabufs, which will mostly be rejected, and mutter will fall back to shm. This behaviour was discussed at guadec last week and hasn't made it into documentation.

I see, I will look into that. Thank you.

grulja:
> Thanks for this! I left some comments on the Gerrit patch set.

Thank you.

columbarius:
> I see, I will look into that. Thank you.

I've implemented this testwise in obs https://github.com/columbarius/obs-studio/tree/egl-modifiers. 
Maybe that's helpfull.

grulja:
@columbarius what is the state of all those changes around modifiers? Has it been already decided it's the way to go? I can see that all your changes have not been merged yet so I wonder whether it's correct to implement them now or there is a chance this will change in the future and I will have to fix WebRTC.

columbarius:
@grulja at guadec i got an ack from @wtay and gnome developers for those changes. Those changes are implemented and tested in the dmabuf patch for the wlroots portal and are looking promising. After guadec i started on the patches for gnome last weekend and since yesterday they are waiting for review. I can't tell how long that will takes. 
I don't know how long the libwebrtc release cicle is, but the additional changes to adapt to the new capability announcements could be done in a second patch after changes in gnome and kde to be safe. The only downside would be that dmabuf screencast could stop working for the time until libwebrtc is adapted.

vchernin:
So I tested [build 54521](https://github.com/flathub/org.chromium.Chromium/pull/135#issuecomment-887736595) a bit more on the integrated amd graphics device (with a proper Fedora install now). Screen sharing _considerably_ smoother with this build than with the stable one (both with the dma-buf experimental option enabled). I just tested [by following my previous instructions](https://github.com/flathub/org.chromium.Chromium/pull/135#issuecomment-887003605) and watching how smooth the preview was.

So while this particular device never seemed to crash on stable, this PR does give a very meaningful screen sharing performance boost. Nice work @grulja (and everyone who's helped)!

I did notice from top cpu usage seems to be a bit higher with this build than with stable. Could this be because this build is "ramping up" more and taking advantage of system resources better? If there's an actual benchmark I could do I could try that out, I have a 4k monitor so that's probably a decent stress test. Maybe there's an easy way to collect data from `radeontop`...

grulja:
> @grulja at guadec i got an ack from @wtay and gnome developers for those changes. Those changes are implemented and tested in the dmabuf patch for the wlroots portal and are looking promising. After guadec i started on the patches for gnome last weekend and since yesterday they are waiting for review. I can't tell how long that will take.
> I don't know how long the libwebrtc release cicle is, but the additional changes to adapt to the new capability announcements could be done in a second patch after changes in gnome and kde to be safe. The only downside would be that dmabuf screencast could stop working for the time until libwebrtc is adapted.

I'm in the process implementing this, I hope you don't mind I take some inspiration. 

I have a question, in https://github.com/columbarius/obs-studio/blob/egl-modifiers/plugins/linux-capture/pipewire.c#L276, shouldn't it return `DRM_FORMAT_RGBA8888`, `DRM_FORMAT_RGBX8888`, `DRM_FORMAT_BGRA8888` or `DRM_FORMAT_BGRX8888` instead of those values you have there? Basically you have the alpha channel at the beginning, while all spa_formats have it at the end.

columbarius:
> I'm in the process implementing this, I hope you don't mind I take some inspiration.

No problem. I'm glad when it's implemented the same way.
 
> I have a question, in https://github.com/columbarius/obs-studio/blob/egl-modifiers/plugins/linux-capture/pipewire.c#L276, shouldn't it return `DRM_FORMAT_RGBA8888`, `DRM_FORMAT_RGBX8888`, `DRM_FORMAT_BGRA8888` or `DRM_FORMAT_BGRX8888` instead of those values you have there? Basically you have the alpha channel at the beginning, while all spa_formats have it at the end.

Good point, I don't know. I've seen both xdpw and obs-studio having the format in inverse order.

I changed it in xdpw and  suddenly i got wrong colours using gstreamer, but it's the same defect i had with the pipewire video-play example, which uses sdl and does the convertion in the right order.

emersion:
Anyone up for contributing a PipeWire page to https://afrantzis.com/pixel-format-guide/?

columbarius:
@grulja opened an issue https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1470

emersion:
@grulja: I've submitted a patch to add docs for PipeWire DMA-BUF negotiation here: https://gitlab.freedesktop.org/pipewire/pipewire/-/merge_requests/845

refi64:
So I'm of course hardly an expert here of course, but I know some tools order the color components in an endian-dependent vs independent matter. For instance, Cairo's color profiles retain their order on any endianness, whereas Skia's profiles do not. Thus, Cairo's ARGB is equivalent to Skia's BGRA on little-endian systems. Could this be a similar scenario?

emersion:
Likely yes, leading to a channel swap. Problem is that the endianess of the PipeWire formats isn't documented AFAIK.

grulja:
> @grulja: I've submitted a patch to add docs for PipeWire DMA-BUF negotiation here: https://gitlab.freedesktop.org/pipewire/pipewire/-/merge_requests/845

Thank you. This was very helpful to understand how this all should work.

I tried to implement it, can you try to test this once the build here finishes? @columbarius @emersion?

As I use KDE Plasma, I'm not able to fully test it as KWin doesn't handle DMA-BUF modifiers over PipeWire, but I will try to write a patch for it. At least I tested whether PipeWire stream is created correctly with all modifiers etc. and it worked when I didn't require DMA-BUF modifiers to be present so it behaved like `DRM_MOD_INVALID` was used, which I assume is the same behavior @columbarius tries to achieve with https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1939.

grulja:
> Build [54830 failed](https://flathub.org/builds/#/builders/32/builds/54830)

Looks like the pach has to be rebased on some older WebRTC version. I will try to do it later.

@columbarius I can also see you sent a MR for KWin so I can test this as well. Thank you.

grulja:
@columbarius @emersion: I can see that `SPA_POD_PROP_FLAG_MANDATORY` is not defined in the PipeWire version available in both the Freedesktop runtime or in the PipeWire version available in Chromium sysroot. Since we need to set DMA-BUF modifiers to be mandatory, what to do about this? I can define this value myself and it will most like work on systems where newer version of PipeWire is running, but do you know what will happen on a system where this is not defined? Will PipeWire just fail or just skip unknown values so only the half of formats where we don't specify modifiers will be announced?

columbarius:
@grulja: Without it pipewire will most likely just ignore that flag and map the format with modifiers against those without.
This can result in wrong negotiation. 

Would it be possible to detect the pipewire version at runtime and avoid announcing modifiers?

grulja:
> @grulja: Without it pipewire will most likely just ignore that flag and map the format with modifiers against those without.
> This can result in wrong negotiation.
> 
> Would it be possible to detect the pipewire version at runtime and avoid announcing modifiers?

Yes, I implemented it now this way. However, you won't be able to test this with this build here as there is apparently older version of PipeWire. Luckily, this is a runtime check so when Chrome is built it is not linked to PipeWire libraries as they are dlopened instead so it should work on newer systems.

vchernin:
> // Modifiers can be used with PipeWire >= 0.3.29

I am no expert on the modifiers or dma-buf itself, but if the PipeWire version provided in the runtime is too old, perhaps you could try adding a more recent PipeWire module to the main manifest (`org.chromium.Chromium.yaml`). That might then produce testable builds.

You could undo part of 9ba724c2c3127f0e5e50b18d5fe3eb9e4ac225a8 and change the version to 0.3.32.

grulja:
I have no idea why, but the official build crashes with SIGILL, while when I tried debug build, it doesn't crash and works just fine. Also my local build (non-flatpak) works just fine.

grulja:
I give up for now, no idea what's wrong and what to try next. The build here crashes with:
```
Thread 1 "chrome" received signal SIGILL, Illegal instruction.
0x000055555f02f844 in webrtc::EglDmaBuf::EglDmaBuf (this=<optimized out>) at ../../third_party/webrtc/modules/desktop_capture/linux/egl_dmabuf.cc:234
234     ../../third_party/webrtc/modules/desktop_capture/linux/egl_dmabuf.cc: No such file or directory.
(gdb) bt
#0  0x000055555f02f844 in webrtc::EglDmaBuf::EglDmaBuf() (this=<optimized out>) at ../../third_party/webrtc/modules/desktop_capture/linux/egl_dmabuf.cc:234
#1  0x000055555c0dae82 in std::__1::make_unique<webrtc::EglDmaBuf>() () at ../../buildtools/third_party/libc++/trunk/include/memory:2006
#2  webrtc::BaseCapturerPipeWire::InitEGL() (this=0x2a6e00481180) at ../../third_party/webrtc/modules/desktop_capture/linux/base_capturer_pipewire.cc:575
#3  0x000055555c0dacbb in webrtc::BaseCapturerPipeWire::OnOpenPipeWireRemoteRequested(_GDBusProxy*, _GAsyncResult*, void*) (proxy=<optimized out>, result=<optimized out>, user_data=0x2a6e00481180)
    at ../../third_party/webrtc/modules/desktop_capture/linux/base_capturer_pipewire.cc:1381
#4  0x00007ffff740fbd9 in g_task_return_now () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#5  0x00007ffff741075d in g_task_return.part () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#6  0x00007ffff747789b in reply_cb () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#7  0x00007ffff740fbd9 in g_task_return_now () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#8  0x00007ffff741075d in g_task_return.part () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#9  0x00007ffff746c3ef in g_dbus_connection_call_done () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#10 0x00007ffff740fbd9 in g_task_return_now () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#11 0x00007ffff740fc1d in complete_in_idle_cb () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#12 0x00007ffff794375f in g_main_context_dispatch () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#13 0x00007ffff7943b08 in g_main_context_iterate.constprop () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#14 0x00007ffff7943bd3 in g_main_context_iteration () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#15 0x000055555ee80ae2 in base::MessagePumpGlib::Run(base::MessagePump::Delegate*) (this=0x2a6e00247e80, delegate=<optimized out>) at ../../base/message_loop/message_pump_glib.cc:400
#16 0x0000555558b043e4 in base::sequence_manager::internal::ThreadControllerWithMessagePumpImpl::Run(bool, base::TimeDelta) (this=0x2a6e0041c340, application_tasks_allowed=true, timeout=...)
    at ../../base/task/sequence_manager/thread_controller_with_message_pump_impl.cc:468
#17 0x0000555558b03b0a in base::RunLoop::Run(base::Location const&) (this=0x2a6e0112e250, location=...) at ../../base/run_loop.cc:132
#18 0x00005555591337ad in content::BrowserMainLoop::RunMainMessageLoop() (this=<optimized out>) at ../../content/browser/browser_main_loop.cc:999
#19 0x0000555559133692 in content::BrowserMainRunnerImpl::Run() (this=0x2a6e00437b50) at ../../content/browser/browser_main_runner_impl.cc:152
#20 0x00005555590816cc in content::BrowserMain(content::MainFunctionParams const&) (parameters=...) at ../../content/browser/browser_main.cc:47
#21 0x000055555907ef12 in content::RunBrowserProcessMain(content::MainFunctionParams const&, content::ContentMainDelegate*) (main_function_params=..., delegate=<optimized out>) at ../../content/app/content_main_runner_impl.cc:597
#22 content::ContentMainRunnerImpl::RunBrowser(content::MainFunctionParams&, bool) (this=0x2a6e00249800, main_params=..., start_minimal_browser=<optimized out>) at ../../content/app/content_main_runner_impl.cc:1080
#23 content::ContentMainRunnerImpl::Run(bool) (this=0x2a6e00249800, start_minimal_browser=false) at ../../content/app/content_main_runner_impl.cc:955
#24 0x0000555558d66560 in content::RunContentProcess(content::ContentMainParams const&, content::ContentMainRunner*) (params=..., content_main_runner=0x2a6e00249800) at ../../content/app/content_main.cc:380
#25 content::ContentMain(content::ContentMainParams const&) (params=...) at ../../content/app/content_main.cc:406
#26 0x000055555ea9f7a2 in ChromeMain(int, char const**) (argc=1, argv=0x7fffffffd498) at ../../chrome/app/chrome_main.cc:151
#27 0x00007ffff664f062 in __libc_start_main () at /usr/lib/x86_64-linux-gnu/libc.so.6
#28 0x000055555e7a28ee in _start () at ../sysdeps/x86_64/start.S:120
```
When I try to print the instruction:
`=> 0x55555f02f844 <webrtc::EglDmaBuf::EglDmaBuf()+852>: ud1 0x2(%eax),%eax
`
When I did a debug build (is_debug = true, is_official_build = false), then it doesn't crash and works just fine. Also my local build (non-flatpak) works just fine.

@refi64 do you have any idea what to try next?

refi64:
Ah, at a glance this looks a *lot* like a [CFI violation](https://www.chromium.org/developers/testing/control-flow-integrity), where unsafe function pointer casts result in a SIGILL. (This happened with an early version of the Flatpak when trying to dynamically check the compiled in FFmpeg codecs.) Seems like the xrandr code [uses `RTC_NO_SANITIZE("cfi-icall")`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/desktop_capture/linux/screen_capturer_x11.cc;l=150?q=path:third_party%2Fwebrtc%20rtc_no_sanitize&ss=chromium%2Fchromium%2Fsrc&start=11) to work around this; try putting it on the EglDmabuf constructor and anywhere else you cast the result of `eglGetProcAddress`.

This doesn't happen on debug builds since those have cfi disabled (but you can enable it for testing).

patches/chromium/PipeWire-capturer-implement-proper-DMA-BUFs-support.patch
```diff
@@ -0,0 +1,935 @@
...
++  RTC_LOG(LS_INFO) << "Egl initialization succeeded";
++  egl_initialized_ = true;
++}
++
++std::unique_ptr<uint8_t[]> EglDmaBuf::ImageFromDmaBuf(int32_t fd,
++                                                      const DesktopSize& size,
++                                                      int32_t stride,
++                                                      uint32_t format,
++                                                      uint32_t offset,
++                                                      uint64_t modifier) {
```
columbarius, line 694: Please extend fd, stride and offset to an array, so we can import buffers with multiple planes.

patches/chromium/PipeWire-capturer-implement-proper-DMA-BUFs-support.patch
```diff
@@ -0,0 +1,935 @@
...
++      SyncDmaBuf(spa_buffer->datas[0].fd, DMA_BUF_SYNC_START);
+     }
++#endif
+ 
+-    src = SPA_MEMBER(map.get(), spaBuffer->datas[0].mapoffset, uint8_t);
++    src = SPA_MEMBER(map.get(), spa_buffer->datas[0].mapoffset, uint8_t);
+ #if PW_CHECK_VERSION(0, 3, 0)
+-  } else if (spaBuffer->datas[0].type == SPA_DATA_MemPtr) {
++  } else if (spa_buffer->datas[0].type == SPA_DATA_DmaBuf) {
++    src_unique_ptr = egl_dmabuf_->ImageFromDmaBuf(
```
columbarius, line 360: spa_buffer->n_datas should give the plane count.

patches/chromium/PipeWire-capturer-implement-proper-DMA-BUFs-support.patch
grulja: Should be now implemented, can you check? One thing I'm not sure about is that later I initialize `src` with:

`src = std::make_unique<uint8_t[]>(strides[0] * size.height());`

Will the stride be same for each plane or should I just get the maximum stride from all planes and use that one?

grulja:
> Ah, at a glance this looks a _lot_ like a [CFI violation](https://www.chromium.org/developers/testing/control-flow-integrity), where unsafe function pointer casts result in a SIGILL. (This happened with an early version of the Flatpak when trying to dynamically check the compiled in FFmpeg codecs.) Seems like the xrandr code [uses `RTC_NO_SANITIZE("cfi-icall")`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/desktop_capture/linux/screen_capturer_x11.cc;l=150?q=path:third_party%2Fwebrtc%20rtc_no_sanitize&ss=chromium%2Fchromium%2Fsrc&start=11) to work around this; try putting it on the EglDmabuf constructor and anywhere else you cast the result of `eglGetProcAddress`.
> 
> This doesn't happen on debug builds since those have cfi disabled (but you can enable it for testing).

Thank you. This actually makes sense and I believe that's the reason for the SIGILL. We will see once the build finishes.

grulja:
Flathub builds seem to fail as there is not enough space. I'll be on vacation starting tomorrow and for the whole next week. I'll try to trigger this build one more time later, but I won't be updating my changes in case something doesn't work. Hopefully this last version will work and people will have a chance to test until I'm back.

refi64:
Seems to be a Flathub issue, as I'm seeing other reports of this in the matrix chat. If this build fails, I'll just re-run it manually once whatever underlying issue is fixed.

vchernin:
If it's possible you might want to bundle the latest [PipeWire 0.3.33](https://gitlab.freedesktop.org/pipewire/pipewire/-/releases/0.3.33) which has improved DMA-BUF negotiation.

columbarius:
> If it's possible you might want to bundle the latest [PipeWire 0.3.33](https://gitlab.freedesktop.org/pipewire/pipewire/-/releases/0.3.33) which has improved DMA-BUF negotiation.

No need to hurry on that, there are still some small details which have to be worked out.

grulja:
bot, build

grulja:
@columbarius finally a build you can try. It works for me at least.

columbarius:
Works for me with implicit modifier as well. Multiplane dmabufs didn't work, because the announced modifiers are different.

columbarius:
> If it's possible you might want to bundle the latest [PipeWire 0.3.33](https://gitlab.freedesktop.org/pipewire/pipewire/-/releases/0.3.33) which has improved DMA-BUF negotiation.

Thinking about this, if we can justify the jump to pipewire 0.3.33 we should do that and adding the `SPA_POD_PROP_FLAG_DONT_FIXATE`  to the modifiers in `buildFormat`. The details missing details are only interesting on the producer side, so the compositor can force the old behaviour by not adding the flag, while libwebrtc would be ready for the compositor to opt into that feature!

grulja:
> > If it's possible you might want to bundle the latest [PipeWire 0.3.33](https://gitlab.freedesktop.org/pipewire/pipewire/-/releases/0.3.33) which has improved DMA-BUF negotiation.
> 
> Thinking about this, if we can justify the jump to pipewire 0.3.33 we should do that and adding the `SPA_POD_PROP_FLAG_DONT_FIXATE` to the modifiers in `buildFormat`. The details missing details are only interesting on the producer side, so the compositor can force the old behaviour by not adding the flag, while libwebrtc would be ready for the compositor to opt into that feature!

Implemented and updated PW to 0.3.33. Thanks you.

PR #151, opened 2021-09-21
21.08 experiments

refi64:


build-aux/bootstrap.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash -e
 
+# Needed to build GN itself.
+. /usr/lib/sdk/llvm12/enable.sh
+
+export LD_LIBRARY_PATH=/usr/lib/sdk/llvm12/lib:$LD_LIBRARY_PATH
```
Erick555, line 6: I can add such thing to enable.sh script if it helps. Does extension path need to be called first or ```LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib/sdk/llvm12/lib``` would work as well?

refi64:
Trying to mix #134 into this to keep all the major upcoming changes together.

PR #155, opened 2021-09-24
Remove x-scheme-handler/ftp from MimeTypes

### LINKED ISSUE #153: Claims to handle  ftp: scheme
`org.chromium.Chromium.desktop` in this repo has `x-scheme-handler/ftp` in the MimeType list, which means it claims to handle the `ftp://` URI scheme.

As best I can tell, Chromium does not support FTP any longer. https://www.chromestatus.com/feature/6246151319715840 claims that FTP support will be fully removed in Chromium 95 (we are at 93) but [this comment](https://bugs.chromium.org/p/chromium/issues/detail?id=333943#c74) claims that the file association was removed for Linux & Mac on 20th May 2021, and certainly the browser does not actually support this scheme. I guess that change has not propagated from the upstream `.desktop` file to the one in this repo.

wjt:
Chromium 91 removed support for FTP.

This line is now identical to that from
https://chromium.googlesource.com/chromium/src/+/main/chrome/installer/linux/common/desktop.template.

Fixes #153

### DIFF
--- a/org.chromium.Chromium.desktop
+++ b/org.chromium.Chromium.desktop
@@ -113,7 +113,7 @@ Icon=org.chromium.Chromium
 Type=Application
 Categories=Network;WebBrowser;
 Keywords=chrome;internet;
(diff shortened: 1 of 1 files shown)


refi64:
Merged by hand, since the CI failure seems to have just been a transient Flathub issue. (Also, nice branch name.)

Issue #158, opened 2021-10-05
Audio renderer error, please restart your computer


markusgod:
Chrome in flatpak fails to open any video and shows error. chrome from repository works fine.
![image](https://user-images.githubusercontent.com/20001291/135975719-71d771c1-8d93-43a1-b911-ddd07361d7e1.png)

Fedora 34, nvidia drivers.

```
--Oct 05 09:51:57 fedora org.chromium.Chromium.desktop[2865]: [2:19:1005/095157.560929:ERROR:chrome_browser_main_extra_parts_metrics.cc(228)] crbug.com/1216328: Checking Bluetooth availability started. Please report if there is no report that this ends.
Oct 05 09:51:57 fedora org.chromium.Chromium.desktop[2865]: [2:19:1005/095157.560954:ERROR:chrome_browser_main_extra_parts_metrics.cc(231)] crbug.com/1216328: Checking Bluetooth availability ended.
Oct 05 09:51:57 fedora org.chromium.Chromium.desktop[2865]: [2:19:1005/095157.560962:ERROR:chrome_browser_main_extra_parts_metrics.cc(234)] crbug.com/1216328: Checking default browser status started. Please report if there is no report that this ends.
Oct 05 09:51:57 fedora org.chromium.Chromium.desktop[2865]: [2:19:1005/095157.567208:ERROR:chrome_browser_main_extra_parts_metrics.cc(238)] crbug.com/1216328: Checking default browser status ended.
Oct 05 09:52:16 fedora org.chromium.Chromium.desktop[3529]: The futex facility returned an unexpected error code.
Oct 05 09:52:16 fedora org.chromium.Chromium.desktop[3543]: Failed to load cookie file from cookie: Permission denied
Oct 05 09:52:16 fedora org.chromium.Chromium.desktop[3543]: The futex facility returned an unexpected error code.
Oct 05 09:52:17 fedora org.chromium.Chromium.desktop[3648]: Failed to load cookie file from cookie: Permission denied
Oct 05 09:52:17 fedora org.chromium.Chromium.desktop[3648]: The futex facility returned an unexpected error code.
```

apawelec:
Same here, after downgrade (`sudo flatpak update --commit=8102b2026553aa2f881e9a44804c2af69e97f38a513323ff5b4cf2ab9774312d org.chromium.Chromium`) it works again.

wjt:
[...]

PR #160, opened 2021-10-05
Back to 20.08

wjt:
This reverts the patches that moved to 21.08, since video playback seems broken with 21.08 (#158).


Erick555:
Since there are many changes unrelated to runtime version maybe downgrading runtime will be sufficient?

refi64:
@Erick555 given the long build times and current broken state, it's easiest to just revert everything now, that way at least people have a functional build available

Issue #164, opened 2021-10-13
Google Meet video streams sporadically freeze


wjt:
In multi-party Google Meet calls, video streams periodically freeze. It seems to happen most frequently when a new stream is added or removed. It affects webcam streams and screenshare streams. It also affects the local playback of my own webcam and screenshare streams – though when it struck my screenshare I was told that others on different browsers could still see updates to the screenshare.

Perhaps a half-dozen people at Endless have seen this on Endless OS. It predates the switch to runtime 21.08. Anecdotally it appears to have struck towards the end of September. This probably means cb9baee9 (93.0.4577.82) was OK and 2e3c01d2 (94) was not.

Google's Chrome package running on the host system is unaffected. I have not tried the Flatpak:ed version from flathub-beta.

wjt:
@dsd, the other colleague I've discussed this with, is actually on Wayland too. I'll try reverting my system to X11 to see if this has any impact.

wjt:
Indeed #165 resolved this for me, I haven't seen it since.

PR #165, opened 2021-10-13
Disable the fieldtrial testing config

refi64:


build-aux/bootstrap.sh
```diff
@@ -32,6 +32,7 @@ tools/gn/bootstrap/bootstrap.py -v --no-clean --gn-gen-args='
     rtc_link_pipewire=true
     rtc_pipewire_version="0.3"
     enable_hangout_services_extension=true
+    disable_fieldtrial_testing_config=true
```
wjt, line 35: Could you explain what the impact is of this change, and why it is desirable, in the commit message?

wjt:
It's early days but since installing this I have indeed not seen #164

PR #201, opened 2022-02-12
Fix Vulkan support in installation steps

refi64:
It seems we were missing some important files for this to be
functional.


refi64:
@liborat could you test this and see if it resolves your WebGL issues?

refi64:
perfect!

PR #228, opened 2022-05-05
Fix crash in PipeWire code when re-negotiating stream parameters

grulja:
This should fix crash in https://github.com/emersion/xdg-desktop-portal-wlr/issues/205.

Upstream change is here: https://webrtc-review.googlesource.com/c/src/+/261300

### DIFF
--- a/patches/all.json
+++ b/patches/all.json
@@ -17,6 +17,7 @@
       "patches/chromium/Clang-build-script-Allow-passing-a-prebuilt-Clang-for-boos.patch",
       "patches/chromium/Clang-build-script-Disable-hwasan.patch",
       "patches/chromium/clang-build-script-Support-disabling-the-bundled-libxml2.patch",
+      "patches/chromium/Webrtc-add-cfi-icall-to-avoid-crash-in-pipewire-functions.patch",
       "patches/ffmpeg/Enable-support-for-libfdk-aac-and-OpenH264.patch",
       "patches/ffmpeg/ffmpeg-Handle-transient-decode-errors-arising-from-libfdk-.patch",
       "patches/ffmpeg/Update-build-configuration.patch"
--- a/patches/chromium/Webrtc-add-cfi-icall-to-avoid-crash-in-pipewire-functions.patch
+++ b/patches/chromium/Webrtc-add-cfi-icall-to-avoid-crash-in-pipewire-functions.patch
@@ -1,0 +1,23 @@
+From 5f2d09cc2338eb75e014badfc037532c561c7d6d Mon Sep 17 00:00:00 2001
+From: Jan Grulich <grulja@gmail.com>
+Date: Thu, 5 May 2022 14:50:34 +0200
+Subject: Add CFI-ICALL to avoid crash in PipeWire functions
+
+We already use RTC_NO_SANITIZE("cfi-icall") for most of the code and
+it looks this one can be triggered recently with pw_loop_signal_event()
+call.
+
+Bug: webrtc:13659
+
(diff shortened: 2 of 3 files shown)


columbarius:
Crashes are gone, but chromium is unable to strip modifiers. This lines are spammed repeatably:
```
[2:166:0506/001704.992303:ERROR:egl_dmabuf.cc(569)] Failed to record frame: Error creating EGLImage - EGL_BAD_ALLOC
[2:166:0506/001704.992330:ERROR:shared_screencast_stream.cc(717)] Dropping DMA-BUF modifier: 72057594037927940 and trying to renegotiate stream parameters
```

columbarius:
Ok bumping the pipewire version fixed it for me. Another difference in the renegotiation code between chromium and obs is this:
```
diff --git a/modules/desktop_capture/linux/wayland/shared_screencast_stream.cc b/modules/desktop_capture/linux/wayland/shared_screencast_stream.cc
index a60b6abcdf..644569fd76 100644
--- a/modules/desktop_capture/linux/wayland/shared_screencast_stream.cc
+++ b/modules/desktop_capture/linux/wayland/shared_screencast_stream.cc
@@ -599,7 +599,7 @@ void SharedScreenCastStreamPrivate::ProcessBuffer(pw_buffer* buffer) {
       RTC_LOG(LS_ERROR) << "Dropping DMA-BUF modifier: " << modifier_
                         << " and trying to renegotiate stream parameters";
 
-      if (pw_client_version_ >= kDropSingleModifierMinVersion) {
+      if (pw_server_version_ >= kDropSingleModifierMinVersion) {
         modifiers_.erase(
             std::remove(modifiers_.begin(), modifiers_.end(), modifier_),
             modifiers_.end());
```

grulja:
> Ok bumping the pipewire version fixed it for me. Another difference in the renegotiation code between chromium and obs is this:
> 
> ```
> diff --git a/modules/desktop_capture/linux/wayland/shared_screencast_stream.cc b/modules/desktop_capture/linux/wayland/shared_screencast_stream.cc
> index a60b6abcdf..644569fd76 100644
> --- a/modules/desktop_capture/linux/wayland/shared_screencast_stream.cc
> +++ b/modules/desktop_capture/linux/wayland/shared_screencast_stream.cc
> @@ -599,7 +599,7 @@ void SharedScreenCastStreamPrivate::ProcessBuffer(pw_buffer* buffer) {
>        RTC_LOG(LS_ERROR) << "Dropping DMA-BUF modifier: " << modifier_
>                          << " and trying to renegotiate stream parameters";
>  
> -      if (pw_client_version_ >= kDropSingleModifierMinVersion) {
> +      if (pw_server_version_ >= kDropSingleModifierMinVersion) {
>          modifiers_.erase(
>              std::remove(modifiers_.begin(), modifiers_.end(), modifier_),
>              modifiers_.end());
> ```

Good catch. Maybe both need to be in version 0.3.40? 

I have this information from here: https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/examples/video-play-fixate.c#L127

columbarius:
Ok looked into it: The reason we need the server to be 0.3.40 is that before it would mark the client after the first round of negotiation as negotiation finished and would start creating buffers ([Issue 1732](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1732)). As such the version of the client is not important at this point. Should fix the example.

grulja:
> Ok looked into it: The reason we need the server to be 0.3.40 is that before it would mark the client after the first round of negotiation as negotiation finished and would start creating buffers ([Issue 1732](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1732)). As such the version of the client is not important at this point. Should fix the example.

Thanks. I opened a MR: https://webrtc-review.googlesource.com/c/src/+/262621,edit

wjt:
Both those upstream requests seem to have landed some months ago – is this patch still needed here?

PR #229, opened 2022-05-10
Bump PipeWire version to 0.3.40

columbarius:
To use the full potential of webrtc screencast we need to build against PipeWire 0.3.40. Until the next Runtime is released we can ship our own version of PipeWire. OBS does the same.

### DIFF
--- a/org.chromium.Chromium.yaml
+++ b/org.chromium.Chromium.yaml
@@ -66,6 +66,38 @@ sdk-extensions:
 modules:
   - libsecret.json
 
+  - name: pipewire
+    buildsystem: meson
+    config-opts:
+      - -Daudiotestsrc=disabled
+      - -Droc=disabled
+      - -Dvideotestsrc=disabled
+      - -Dvolume=disabled
+      - -Dvulkan=disabled
+      - -Ddocs=disabled
+      - -Dman=disabled
+      - -Dbluez5-codec-ldac=disabled
+      - -Dbluez5-codec-aptx=disabled
+      - -Dlibcamera=disabled
+      - -Dudevrulesdir=/app/lib/udev/rules.d/
+      - -Dsession-managers=[]
+      - -Dtests=disabled
+      - -Dexamples=disabled
+      - -Dpw-cat=disabled
+    cleanup:
+      - /include
+      - /bin
+    sources:
+      - type: git
+        url: https://github.com/pipewire/pipewire.git
+        tag: 0.3.40
+        commit: 7afd80052b7c49754a13c9ab49c368f95b60e0a7
+        x-checker-data:
+          type: anitya
+          project-id: 57357
+          stable-only: true
+          tag-template: $version
+
   - name: gtk-cups-backend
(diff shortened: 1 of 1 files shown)


org.chromium.Chromium.yaml
```diff
@@ -64,6 +64,8 @@ sdk-extensions:
   - org.freedesktop.Sdk.Extension.openjdk11
 
 modules:
+  - pipewire.json
```
refi64, line 67: Can the body of this just be inline (& right below `libsecret.json`)?

pipewire.json
```diff
@@ -0,0 +1,33 @@
...
+  ],
+  "cleanup": [
+    "/include",
+    "/bin"
+  ],
+  "sources": [
+    {
+      "type": "git",
+      "url": "https://github.com/pipewire/pipewire.git",
+      "tag": "0.3.40",
```
refi64, line 29: This should be wired up to [the external data checker](https://github.com/flathub/flatpak-external-data-checker), otherwise it will almost definitely fall behind on bug fix releases and such.

columbarius:
I've updated this since at least PipeWire 0.3.40 is required to get screencast working with explicit modifiers supported by the wlroots portal. Otherwise chromium is unable to drop unsupported modifiers correctly.

org.chromium.Chromium.yaml
columbarius: inlined this.

pipewire.json
columbarius: I've added the x-checker-data block. is there a way to confirm i did it correctly?

tinywrkb: You can test it by setting an old version/tag for the module and running f-e-d-c (flatpak-external-...). The quickest way to get f-e-d-c is to [install it from Flathub](https://github.com/flathub/flatpak-external-data-checker/#installation).

Don't use the git checker, AFAIK it has an issue with fetching timestamp, and the end result is that it will open an update PR every time it runs if the previous one wasn't merged and applied the changes.

You can use Anitya checker like this
```
"x-checker-data": {
    "type": "anitya",
    "project-id": 57357,
    "stable-only": true,
    "tag-template": "$version"
}
```
You can also use JSON checker with GitLab's API
```
"x-checker-data": {
    "type": "json",
    "url": "https://gitlab.freedesktop.org/api/v4/projects/pipewire%2Fpipewire/repository/tags",
    "tag-query": "first(.[].name) ",
    "version-query": "$tag | sub(\"^v\"; \"\")",
    "commit-query": ".[] | select(.name==$tag) | .commit.id",
    "timestamp-query": ".[] | select(.name==$tag) | .commit.created_at"
}
```

columbarius: Thanks. Using anitya now and tested with f-e-d-c.

columbarius:
Btw. should I add in the commit message that this should be reverted after the runtime is updated to 22.08?

columbarius:
Closing this since the runtime was updated.

PR #247, opened 2022-09-30
desktop: Add "google" keyword

wjt:
I'm watching a usability testing session. The user is struggling to find a web browser, which they refer to as "google". The computer has Chromium & Firefox installed, but not Chrome. So they cannot find the browser, and are trying to install a third one.

Chrome (stable and dev) show up in the search in Software, as does ungoogled-chromium. But this app doesn't mention the word "google" in its desktop file or metainfo, despite being more googly than ungoogled-chromium.

Add "google" as a keyword.


refi64:
Nice catch.

PR #243, opened 2022-09-12
Update to runtime 22.08

columbarius:
22.08 brings a newer PipeWire version and might make #229  obsolete.

PR #254, opened 2022-10-28
Update to runtime 22.08 [take 2]

refi64:
Supersedes #243, @columbarius I kept authorship information in your commit.

The main change is that includes migrating to `org.chromium.Chromium.BaseApp`, which should have a newer krb5 that's compatible with OpenSSL 3.


Erick555:
[This](https://github.com/flathub/org.chromium.Chromium/blob/4046ef821b7fd5ddc657736251831a990337829f/build-aux/bootstrap.sh#L21-L22) needs to be updates as well. Maybe create variable like LLVM_VERSION=14 and call it everywhere.

refi64:
Really I should just like, use whatever LLVM SDK extension is available, there will only ever be one...but I'll leave that cleanup for another day :sweat_smile:

joshua-stone:
The latest Chromium test build still seems to have [the cursor issue that affected Electron apps running in Wayland](https://github.com/flathub/org.signal.Signal/issues/349) even though it's using the newer 22..08 runtime that had the cursor fix.

I don't think that issue should be a blocker for this change though.

PR #255, opened 2022-10-30
Allow hardware acceleration

Samisafool:


Samisafool:
Oops

But it works when flatseal overrides... Weird

Erick555:
Manifest contains `--device=all` so `--device=dri` doesn't do anything unless you removed the former manually.

refi64:
Indeed, this shouldn't make a practical difference for hardware acceleration (which is just universally messed up, I've been trying to look into it but it's not clear).

PR #274, opened 2023-02-24
Change socket=x11 to socket=fallback-x11

violetmage:
Allows systems with wayland available to enable only the wayland socket, and possibly avoid showing a "This app uses a legacy window system" warning.

### DIFF
--- a/org.chromium.Chromium.yaml
+++ b/org.chromium.Chromium.yaml
@@ -16,7 +16,7 @@ finish-args:
   - --share=network
   - --socket=cups
   - --socket=pulseaudio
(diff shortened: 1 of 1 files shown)


Erick555:
Does chromium run on native wayland by default without additional settings now? In the past it didn't which would make it broken with this change.

violetmage:
On my systems, it appears so, however I have not done any proper testing at the moment.

violetmage:
3am me is not smart.

Noticed that the chromium flatpak had ozone built and assumed it was enabled by default.

My mistake

PR #282, opened 2023-04-19
Test EGL fallback fixes

refi64:


refi64:
These are in already.

PR #298, opened 2023-06-18
patches: Add VAAPI wayland patches

EliasOfWaffle:
Patches from: https://aur.archlinux.org/cgit/aur.git/plain/vaapi-add-av1-support.patch?h=chromium-wayland-vaapi


EliasOfWaffle:
Please some dev can stop the build if is possible?
Sorry for my big mistake

Issue #187, opened 2021-12-29
Consider carrying the patches from chromium-wayland-vaapi for hardware accelerated decode under Wayland


jbalme:
There are a set of patches that appear to enable hardware-accelerated video decoding on Wayland used in the following AUR package:

https://aur.archlinux.org/packages/chromium-wayland-vaapi/

I'm unaware of the origins of these patches, but it might be something to look into for a Wayland-first experience.

refi64:
I'm a bit hesitant to include these for two main reasons:

- Every patch we add is something that needs to be merged with new releases. I don't want to be bottlenecked on updates waiting for an external entity, but I don't know enough about VA-API internals to be rebasing them myself.
- The patch seems a bit...simplistic, especially with the randomly changed magic numbers (e.g. `+ 1` turning into `+ 3`) and given that I believe Igalia had mentioned making fully Ozone-complaint VA-API would take some internal changes. Maybe this is just a bit young and I'm overthinking it, but VA-API is definitely the type of area where individual hardware can make a huge difference, and I don't want to end up with some patches that only work in some odd subset of cases.

Now, once a patch is merged in, I'd probably backport it to a current release, but I'm not keen to keep it in otherwise atm.

PR #299, opened 2023-06-19
Wayland: Build with VAAPI wayland patches

EliasOfWaffle:
The main ideia is for build using vaapi patches from aur https://aur.archlinux.org/packages/chromium-wayland-vaapi

And sorry for my old pull request.

### DIFF
--- a/patches/all.json
+++ b/patches/all.json
@@ -2,6 +2,8 @@
   {
     "type": "patch",
     "paths": [
+      "patches/chromium/0001-ozone-wayland-add-VA-API-support.patch",
+      "patches/chromium/vaapi-add-av1-support.patch",
       "patches/chromium/flatpak-Add-initial-sandbox-support.patch",
       "patches/chromium/flatpak-Expose-Widevine-into-the-sandbox.patch",
       "patches/chromium/flatpak-Adjust-paths-for-the-sandbox.patch",
--- a/patches/chromium/0001-ozone-wayland-add-VA-API-support.patch
+++ b/patches/chromium/0001-ozone-wayland-add-VA-API-support.patch
@@ -1,0 +1,128 @@
+From: Maksim Sisov <msisov@igalia.com>
+Date: Wed, 20 Jan 2021 09:50:22 +0200
+Subject: [PATCH] ozone/wayland: add VA-API support.
+
+This patch ads VA-API support utilizing old VA-API path used for
+ChromeOS, which can be buggy on some devices (currently tested
+on Intel Gen8 and Gen9 with Gen8 having some minor bugs).
+
+It's known that a new VA-API is being developed atm and once it's ready,
+we will switch to a new path, which should be more stable.
+
+Upstream-Status: Inappropriate
+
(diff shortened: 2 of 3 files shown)


refi64:
This has [come up before](https://github.com/flathub/org.chromium.Chromium/issues/187), and unfortunately I still don't think this is something that belongs here. It's a multi-year old patch that was already outdated when created and was known to be a bit buggy at the time (both of these were outlined in the commit message). I'm not familiar enough with this code to be able to debug and rebase it on a new releases.

Chromium *does* also support this nowadays, just in a somewhat roundabout way (it goes through Vulkan). I have absolutely zero idea how that works, but I have equally zero idea how the old path that this patch uses works too.

EliasOfWaffle:
> This has [come up before](https://github.com/flathub/org.chromium.Chromium/issues/187), and unfortunately I still don't think this is something that belongs here. It's a multi-year old patch that was already outdated when created and was known to be a bit buggy at the time (both of these were outlined in the commit message). I'm not familiar enough with this code to be able to debug and rebase it on a new releases.
> 
> Chromium _does_ also support this nowadays, just in a somewhat roundabout way (it goes through Vulkan). I have absolutely zero idea how that works, but I have equally zero idea how the old path that this patch uses works too.

Sorry but this patch not need to be adapt, the source from chromium wayland aur is building, this patch is working in a test, and the vaapi works now with new VaapiDecodeLinuxGL this is the way to get vaapi in OpenGL 
The CI/CD have builded and this is a proof from a user that had tested this artifact build, but had problems with older intel gpus generations, this not working for some whatever reason


7f3c5e65-7a8c-4103-81c0-60578522f2b4

EliasOfWaffle:
@refi64

EliasOfWaffle:
For run is necessary this flags 
--enable-features=VaapiVideoDecodeLinuxGL
--ozone-platform-hint=wayland

EliasOfWaffle:
@refi64

tankaras:
@refi64 Please re-open and merge. Chromium team don't care about their implementation of vaapi. It has had a memory leak [bug](https://gitlab.freedesktop.org/mesa/mesa/-/issues/9220) for a while now under Xwayland with the flag `--enable-features=VaapiVideoDecodeLinuxGL`. 

The wayland patch fixes this problem until Chromium can support vaapi on native Wayland. Have tested with the above mentioned AUR and have no memory leaks.

EliasOfWaffle:
> 

Is better to i create a new merge request or waiting the @refi64 decision ?

EliasOfWaffle:
> @refi64 Please re-open and merge. Chromium team don't care about their implementation of vaapi. It has had a memory leak [bug](https://gitlab.freedesktop.org/mesa/mesa/-/issues/9220) for a while now under Xwayland with the flag `--enable-features=VaapiVideoDecodeLinuxGL`.
> 
> The wayland patch fixes this problem until Chromium can support vaapi on native Wayland. Have tested with the above mentioned AUR and have no memory leaks.

Is better to submit this a new merge request or waiting @refi64 ?

PR #321, opened 2023-09-16
allow Widevine on all architectures if present

theofficialgman:
refer to https://bugs.launchpad.net/ubuntu/bionic/+source/chromium-browser/+bug/2008433

Feel free to add yourself if it easier to maintain @refi64 
This will allow the use of widevine on all architectures if it is found, not just x64.
This was added to all Ubuntu chromium (bionic and snap) back in April.


refi64:
For simplicity I'm indeed folding it into #329 but I did point back to this PR from the commit message there.

PR #327, opened 2023-10-09
Draft: Update Freedesktop SDK to 23.08 and Chromium base app to 23.08

EliasOfWaffle:
Depends on https://github.com/flathub/org.chromium.Chromium.BaseApp/pull/31


EliasOfWaffle:
Future commits: Update SDK modules including LLVM

PR #353, opened 2024-02-02
Make UnityLauncherAPI work

JakobDev:
This allows showing the download progress in the Window Icon.

grafik

It requires libunity, as you can see [here](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/download/download_status_updater_linux.cc;l=76;drc=153865a15b33852de58d6356625be98e6266202c).

While this API is also supported outside Unity/KDE, Chromium does check if it's running on Unity/KDE, we may need to set `XDG_CURRENT_DESKTOP=KDE` to make it work on other Desktops too.

### DIFF
--- a/chromium.sh
+++ b/chromium.sh
@@ -31,5 +31,6 @@ fi
 # Unlink any profiles from the sync keys to avoid any expected deletions.
 unlink_profiles
 
+export CHROME_DESKTOP="org.chromium.Chromium.desktop"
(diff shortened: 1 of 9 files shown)


rany2:
Can this be added to the BaseApp first?

JakobDev:
You mean libunity? I can do it, but 'm not sure if any other Browser makes use of it.

PR #369, opened 2024-03-09
yml: Set VK Driver and Layers Env variables

EliasTheGrandMasterOfMistakes:
VK_ADD_LAYER_PATH defines where is path who contains the vk layers. VK_ADD_DRIVER_FILES defines where can find vulkan drivers icd.d files.

This can be use in future for others flathub chromium's based packages

### DIFF
--- a/org.chromium.Chromium.yaml
+++ b/org.chromium.Chromium.yaml
@@ -11,6 +11,8 @@ finish-args:
   - --filesystem=/run/.heim_org.h5l.kcm-socket
   - --device=all
   - --env=LD_LIBRARY_PATH=/app/chromium/nonfree-codecs/lib
+  - --env=VK_ADD_LAYER_PATH=/usr/lib/x86_64-linux-gnu/GL/vulkan
+  - --env=VK_ADD_DRIVER_FILES=/usr/lib/x86_64-linux-gnu/GL/vulkan/icd.d
   - --share=ipc
   - --share=network
   - --socket=cups


EliasTheGrandMasterOfMistakes:
someone can cancel the build? is not necessary is just a env var, you can set in flatpak and test in your actual browser.

This fixes vk icd search on Angle Vulkan Backend and allow to use Vulkan Backend without problems

@ebassi @refi64

Erick555:
If you expect this to be merged then it must pass the build. BTW: should the fix be done on [runtime level](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/release/23.08/elements/flatpak-images/platform.bst#L19)?

EliasTheGrandMasterOfMistakes:
> If you expect this to be merged then it must pass the build. BTW: should the fix be done on [runtime level](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/release/23.08/elements/flatpak-images/platform.bst#L19)?

Yes is just a env var, is possible to define this two env var in your chromium and testing, this allow to chromium uses the layers and icds from drivers, this fix the problem about vulkan drivers not be found by chromium and allow to use angle vulkan renderer and vulkan capabilities

EliasTheGrandMasterOfMistakes:
On my system this fix problems nothing Vulkan ICD is detected by vulkan loader.
IDK if this affects others users, but ungoogledchromium looks have a issue about it

EliasTheGrandMasterOfMistakes:
Sorry others people have confirmed this bug in a telegram group exist in chromium and ungoogled chromium, if is necessary i will send this pr first to ungoogled chromium for a test

rany2:
We also build for ARM64, this has be done in some other way.

refi64:
Yeah hardcoding it as x64 isn't gonna really work here.

PR #370, opened 2024-03-09
cobalt: Enable Chrome Refresh Features

EliasTheGrandMasterOfMistakes:
It's allow to have a new Chrome UI based that uses Material UI design

### DIFF
--- a/cobalt.ini
+++ b/cobalt.ini
@@ -3,3 +3,7 @@ EntryPoint=/app/chromium/chrome
 
 [DefaultFeatures]
 Enabled=WebRTCPipeWireCapturer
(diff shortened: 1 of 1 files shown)


Erick555:
Does it allows or force it? When chromium team considers it ready then they will enable it by default. Flatpak chromium isn't should tweak arbitrary settings.

EliasTheGrandMasterOfMistakes:
> team

It's enable and force.

Chrome looks just use it, i think this will never or leave amount of time to be used as default.

refi64:
We don't really want to change any defaults like this ahead of upstream.

PR #376, opened 2024-03-20
Make vulkan driver visible to chrome(ium)

rainDiX:
I was trying to play with WebGPU in chromium and noticed that anything vulkan was not working. Thankfully, I stumbled upon this comment https://github.com/flatpak/flatpak/issues/5131#issuecomment-1947245971

TL;DR chrome(ium) vendors libvulkan, so it does not contains the patch that search the correct path for drivers, so this patch set the `VK_DRIVER_FILES` variable.

I think this should also be applied to the other chromium-based browser on flathub, so far I tested chromium and ungoogled-chromium.

### DIFF
--- a/chromium.sh
+++ b/chromium.sh
@@ -32,4 +32,5 @@ fi
 unlink_profiles
 
 export LIBGL_DRIVERS_PATH=/usr/lib/$(uname -m)-linux-gnu/GL/lib/dri
+export VK_DRIVER_FILES=/usr/lib/$(uname -m)-linux-gnu/GL/vulkan/icd.d
 exec cobalt "$@"


rainDiX:
I just saw #369, and I think this approach is better since it also works on ARM

refi64:
If this is something that other Chrome variants can also use, it might be better inside [cobalt](https://github.com/refi64/cobalt) which is what's responsible for starting all those browsers. In particular it should be possible to largely copy-paste the [existing code that does this for the libgl drivers path](https://github.com/refi64/cobalt/blob/99fd12a318acd3d169ab491a051353258fd0e72a/src/cobalt-launcher.c#L272-L274) (setting this shouldn't be needed in the startup script here too, I think I just forgot to clean that up?). If you want you can try to PR it, or I can just push a commit for it and link back to this PR for credit purposes?

rainDiX:
@refi64 you can push a commit that's fine by me.

PR #392, opened 2024-04-29
fix access to kwallet on KDE Plasma6, fix #391

### LINKED ISSUE #391: No access to kwallet on KDE Plasma 6
After my upgrade to Fedora 40 with KDE Plasma 6, chromium seem to have lost access to kwallet, even tho `org.freedesktop.secrets` session buss name should still allow it. Adding `org.kde.kwalletd6` seem to fix the problem.

Szwendacz99:
After my upgrade to Fedora 40 with KDE Plasma 6, chromium seem to have lost access to kwallet, even tho `org.freedesktop.secrets` session buss name should still allow it. 

Adding `org.kde.kwalletd6` seem to fix the problem.

Fixes #391

### DIFF
--- a/org.chromium.Chromium.yaml
+++ b/org.chromium.Chromium.yaml
@@ -26,6 +26,7 @@ finish-args:
   - --talk-name=org.freedesktop.ScreenSaver
   - --talk-name=org.freedesktop.secrets
   - --talk-name=org.kde.kwalletd5
+  - --talk-name=org.kde.kwalletd6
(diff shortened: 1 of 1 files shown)


bbhtt:
The faliure was caused by the recent fdsdk fiasco. I can merge it without building if you want.

wjt:
@bbhtt sounds good, that would save a few electrons from redundantly building a whole browser.

PR #402, opened 2024-05-29
Bundle rust nightly

bbhtt:


bbhtt:
I suggest keeping x-checker disabled to avoid too many PRs.

bbhtt:
It's getting stripped...

bbhtt:
https://github.com/chromium/chromium/blob/9025a06bf980059476f678e27f4559428316436c/build/config/rust.gni#L101 the toolchain_has_rust check is somehow failing. Maybe this is a matter of setting the correct library paths.

bbhtt:
@refi64 please take a look at this, rust-nightly is now being converted to extra-data. Once that is published, chromium builds will fail due to the change in library paths and append-path.

refi64:
oof, is there a timeline for that? I was hoping to sync the version with chromium's instead of leaving it stuck via an external-data-checker change but I hadn't gotten around to it yet.

bbhtt:
It's switched already.

refi64:
Welp guess wiring up the external data checker is future me's problem.

rany2:
> Welp guess wiring up the external data checker is future me's problem.

This is how I'm doing it in UGC: https://github.com/flathub/io.github.ungoogled_software.ungoogled_chromium/blob/763f4e4bb5570fd42300de5723303ac838c18e5d/maint/update_version.sh#L66

refi64:
@bbhtt quick note, github never sends email updates when an issue body is edited, so I actually never noticed you added the background details into the original issue. perhaps useful to just add a separate comment for any major changes :sweat_smile:

Issue #362, opened 2024-02-22
Give Chromium access to xdg-download


thatLeaflet:
The Chromium flatpak by default has access to the entire home folder. On most Linux system, this works fine as most people have their download folders in their home folder.

However, some people, like me, have their XDG download folder on secondary drives. When attempting to download anything, the download operation silently fails (as in Chromium doesn't report any issues, but the file isn't anywhere in the filesystem).

This can be worked around with using Flatseal to grant Chromium access to xdg-download, but it would be better if it had this access by default.

thatLeaflet:
Yes, although the xdg-download path is the most important. I haven't run into any issues with the other xdg paths because the file picker portal can handle those.

AndreiSva:
To add to this, I think giving Chromium access to the whole home directory is a bit extreme and potentially bad for security.

TrusterX:
i second that - i removed access to home.

I wouldn't add any other directory, as uploads are handled through the Portal anyway, so there is no need to add them, i guess. And if i have to test a local website, i just copy it in the download directory... or you create a designated folder for this use case and give accress to Chromium (for eg. ~/.local/www/)

PR #495, opened 2025-05-18
Replace full home acess with xdg-download

andreldmonteiro:
Closes issue #362


andreldmonteiro:
Hi @refi64 can we get this pull in?

refi64:
This isn't really viable because it's a rather large breaking change: this flatpak *predates* the directory chooser portal's existence, and thus there are a non-zero amount of Chromium userdata dirs lying around that contain full storage paths in various places. e.g. if someone changed the downloads directory, then this change would silently break that, and downloads would still occur but just end up in a hidden place.

andreldmonteiro:
I thought of the changing of the download directory of course but if you think about it that should probably be very rare that someone did that, the current full access is dangerous for a browser, now that the change can be made, it should be made. 

The official Firefox Flatpak from Mozilla barely has more access than xdg-download.

andreldmonteiro:
This pull request can be enhanced by adding more filesystem access to align with the ungoogled chromium flatpak that is more restricted by default and so they have already figured things that are needed like:

  - --filesystem=/run/.heim_org.h5l.kcm-socket
  - --filesystem=xdg-run/pipewire-0
  # To install a PWA application
  - --filesystem=home/.local/share/applications:create
  - --filesystem=home/.local/share/icons:create
  # To allow installing shortcuts on the desktop
  - --filesystem=xdg-desktop
  # For default download directory to work as expected
  - --filesystem=xdg-download
  # For MPRIS media cover art
  - --filesystem=/tmp
  # For GNOME proxy resolution
  - --filesystem=xdg-run/dconf
  - --filesystem=~/.config/dconf:ro
  # For KDE proxy resolution (KDE5 only)
  - --filesystem=~/.config/kioslaverc

The main new one is the access to the desktop.

PR #587, opened 2026-05-11
Don't use the proxy portal by default

### LINKED ISSUE #50: no proxy extension is working

### LINKED ISSUE #173: No way to set proxy. Global proxy settings do not work
Can't change proxy settings in chromium. Global settings from KDE don't work either.
There must be a way to set proxy

### LINKED ISSUE #199: Unable to use Proxy Auto Configuration / PAC file via system or extension
:wave: I need to use a [PAC file](https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling/Proxy_Auto-Configuration_PAC_file) to access a certain network and am unable to do so using the flathub-provided version of Chromium.

I've tried setting the proxy via the system (I'm using gnome) as well as leveraging a "PAC URL" or a manually-entered PAC script in plugins like Proxy Switchyomega:

|System settings|Switchyomega|
|---|---|
|Screenshot from 2022-02-05 10-47-55-smudged|Screenshot from 2022-02-05 11-05-37|

In both cases, the flathub-provided version of Chromium seems to ignore the PAC configuration on matching domain names and always return a direct connection. It seems like there may be some internal-to-chromium DNS resolution going on that bypasses the proxy setting.

The Switchyplus docs have some [mention of DNS prefetching](https://github.com/FelisCatus/SwitchyOmega/wiki/DNS-and-SOCKS-proxy) causing a similar effect, but none of the workarounds seem to work for me.

If I configure either the system proxy or switchyplus to __always__ use the proxy (e.g. not using a PAC file), everything works as expected and __all__ requests are routed through the remote host.

Screenshot from 2022-02-05 10-48-30

### Non-flathub Chromium

PAC files work as expected using both the system config & via extension on my distro-provided (Fedora 35) version. That said, it's a couple major versions behind flathub (96 vs 98), so not a direct comparison.

### Related Issue(s)

It sounds like this is probably related to #173, but I thought it worth spelling out in a new issue just in case it's not.

### LINKED ISSUE #355: --proxy-pac-url option is ignored
### Description
 
`--proxy-pac-url=http://wpad/wpad.dat` is ignored.

### How to Reproduce?

1. verify via `curl http://wpad/wpad.dat` that proxy autoconfig data is valid.
2. launch `flatpak run org.chromium.Chromium  --proxy-pac-url=http://wpad/wpad.dat` 
3. open `http://ip4.me` or any other similar site that displays IP address

### Actual behaviour

Shows actual IP address.

### Expected behaviour
 
Shows proxy IP address.

### Relevant log output

Repeat same steps using `/usr/bin/chromium-browser --proxy-pac-url=http://wpad/wpad.dat` and observe that proxy IP is returned.

### Notes

- I initially suspected that this was occurring because `wpad` was failing to resolve, however isn't the case as the same problem occurs when substituting `wpad` with the proxy server's IP.
- The original thread is available at https://github.com/flathub/com.github.Eloston.UngoogledChromium/issues/129
- Related thread: https://github.com/flathub/com.github.Eloston.UngoogledChromium/issues/129

refi64:
The current implementation can't fall back to PAC scripts (the portal doesn't even provide that, and it probably shouldn't), and there's no way to disable it if unused. Perhaps it could explicitly use the V8 resolver and then only fallback to the system one, but that would take a non-trivial amount of wiring. Instead, just flip it off for now, unless the user passes `--use-system-proxy-resolver` (the same mechanism used for the newer Windows implementation).

Fixes: #50
Fixes: #173
Fixes: #199
Fixes: #355

### DIFF
--- a/patches/chromium/Add-support-for-respecting-system-proxy-settings-when-runn.patch
+++ b/patches/chromium/Add-support-for-respecting-system-proxy-settings-when-runn.patch
@@ -1,4 +1,4 @@
-From 8eb17b876489bc1809ac902d0e75ebd2cea72e9f Mon Sep 17 00:00:00 2001
+From 7b554fb9aae32b623de48336e3cc564507cae492 Mon Sep 17 00:00:00 2001
 From: Andre Moreira Magalhaes <andre.magalhaes@endlessos.org>
 Date: Thu, 17 Dec 2020 19:05:10 -0300
 Subject: [PATCH] Add support for respecting system proxy settings when running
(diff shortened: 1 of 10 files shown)


refi64:
bot, build

refi64:
bot, build

PR #602, opened 2026-07-11
Allow talking to org.a11y.Bus for screen reader support

harryg02:
Screen readers such as Orca receive no output from the Chromium Flatpak because the sandbox does not permit talking to the accessibility bus name on the session bus. 

Add --talk-name=org.a11y.Bus to finish-args. This matches the fix already shipped by other Flatpak apps (e.g. OBS Studio, [obsproject/obs-studio#5869](https://github.com/obsproject/obs-studio/pull/5869)). This is a Flatpak sandbox permission in the manifest, not an application code patch.

### Testing:
With Orca running, verified that speech works after adding the permission (`flatpak run --talk-name=org.a11y.Bus org.chromium.Chromium`) and is silent with `--no-talk-name=org.a11y.Bus.`

### DIFF
--- a/org.chromium.Chromium.yaml
+++ b/org.chromium.Chromium.yaml
@@ -21,6 +21,7 @@ finish-args:
   - --system-talk-name=org.freedesktop.Avahi
   - --system-talk-name=org.freedesktop.UPower
   - --talk-name=com.canonical.AppMenu.Registrar
+  - --talk-name=org.a11y.Bus
   - --talk-name=org.freedesktop.FileManager1
   - --talk-name=org.freedesktop.Notifications
   - --talk-name=org.freedesktop.ScreenSaver


Erick555:
In general accessibility bus service is [proxied by flatpak](https://github.com/flathub-infra/flatpak/blob/fdb1741789329f2cc7e25fb53ecf4fa5c12afe34/common/flatpak-run-dbus.c#L380-L485) so direct access shouldn't be needed by most apps.  For example Firefox is one of the apps which use a11y in [non-standard way](https://bugzilla.mozilla.org/show_bug.cgi?id=1756369) which prevent proxy from working.

I didn't saw anyone reporting this for chromium before and I don't see any chromium/electron app using your workaround in flathub. Do you have same issue for any other flatpak app or just chromium?

> This is a Flatpak sandbox permission in the manifest, not an application code patch.

If something in chromium codebase prevents using proxy then code patch may be even better way to fix it if it's not too complicated change. Direct dbus access allows escaping sandbox unfortunately.

harryg02:
> In general accessibility bus service is [proxied by flatpak](https://github.com/flathub-infra/flatpak/blob/fdb1741789329f2cc7e25fb53ecf4fa5c12afe34/common/flatpak-run-dbus.c#L380-L485) so direct access shouldn't be needed by most apps. For example Firefox is one of the apps which use a11y in [non-standard way](https://bugzilla.mozilla.org/show_bug.cgi?id=1756369) which prevent proxy from working.
> 
> I didn't saw anyone reporting this for chromium before and I don't see any chromium/electron app using your workaround in flathub. Do you have same issue for any other flatpak app or just chromium?
> 
> > This is a Flatpak sandbox permission in the manifest, not an application code patch.
> 
> If something in chromium codebase prevents using proxy then code patch may be even better way to fix it if it's not too complicated change. Direct dbus access allows escaping sandbox unfortunately.

On "is it just Chromium": I tested Epiphany (GNOME Web) as a Flatpak, it has no org.a11y.Bus entry and the screen reader still reaches it. Among the browsers I tested, Chromium is the only one that needs the flag manually added to be reachable. 

For contrast, Firefox already ships with org.a11y.Bus allowed by default (listed under dbus access in [Mozilla's Flatpak packaging docs](https://firefox-source-docs.mozilla.org/build/buildsystem/flatpak.html)), with an open request to drop it once Firefox no longer needs direct bus access ([bug 1756369](https://bugzilla.mozilla.org/show_bug.cgi?id=1756369)). I confirmed the dependency is real and identical in behavior: 
`flatpak run --no-talk-name=org.a11y.Bus org.mozilla.firefox` makes Orca unable to reach Firefox content, the same way Chromium is silent without the permission. So this PR would just bring Chromium in line with an already-accepted browser that currently needs the same permission.

On the security tradeoff, I agree granting direct org.a11y.Bus access widens the sandbox, and I don't want to push a workaround if the real fix belongs in Chromium. I don't know Chromium's a11y internals well enough to say why it bypasses the proxy when Epiphany doesn't, so I won't speculate about the mechanism. I agree the cleaner end state is Chromium not needing it at all.

You can reproduce both cases with the steps I posted earlier (My first comment) and the Firefox command above. Happy to hold or close this in favor of an upstream fix if you think that's the better route, or keep it as an interim only if you'd consider that acceptable.

PR #627, opened 2026-09-16
Use a custom workflow+script to update to new Chromium releases

refi64:
The upstream Chromium tarballs are, *again*, lagging behind. It took several days for tarballs for 153 to be uploaded, and as of right now the tarballs for the latest bugfix on top are still not available, despite the release going out *yesterday*. Meanwhile, the chromium-linux-tarballs project originally done by the Gentoo Chromium maintainers has had the tarballs up for hours. This has happened several times already, and it's not really worth trying to bounce back and forth between URLs to use as fallback.

Of course, this is yet *another* change to the sources used that needs to be wired up to update automation. New Chromium releases are adding more and more bespoke things that need to be updated in sync with the official releases, and wiring it up all properly to the external data checker is an effort I don't really want to undertake. Thus, this *also* switches to a custom nushell script to actually handle the update process + a GitHub Actions workflow to check for updates on a schedule, which makes it easier to automatically update esbuild, bindgen, tsc, etc. (This of course isn't really an issue with the external data checker *itself*, per se; it's just that the update frequency here is so high that it's hard to keep up from a separate project, rather than just being inline in this repository.)

This does *not* wire up the nightly rust versions, because the use of a non-chromium rust toolchain is a ticking time bomb that I'll have to resolve separately.


refi64:
(this is draft because I expect I probably broke something incidentally)

refi64:
bot, build

refi64:
bot, build

refi64:
bot, build

